PR #2, opened 2018-10-18
Release/typeform connector new api

michaloo:


packages/connectors/hull-outreach/server/actions/oauth.js
```diff
@@ -45,7 +44,7 @@ function oAuthAction(deps: Object) {
     onAuthorize: req => {
       const { refreshToken, params } = req.account || {};
       const { access_token } = params || {};
-      return settingsUpdate(req.hull, {
+      return req.hull.helpers.settingsUpdate({
```
timliuhull: wait, I thought settingsUpdate wasn't going to be in the helpers anymore in 0.14...

packages/connectors/hull-outreach/test/integration/helper/connector-mock.js
```diff
@@ -35,11 +35,11 @@ class ContextMock {
       })
     };
     this.helpers = {
-      updateSettings: () => {
+      settingsUpdate: () => {
```
timliuhull: I updated this on my branch to pass tests, but using the settingsUpdate method instead of the one from helpers...

packages/hull/test/unit/helpers/settings-update-test.js
```diff
@@ -3,7 +3,7 @@ const { expect } = require("chai");
 const sinon = require("sinon");
 const Promise = require("bluebird");
 
-const settingsUpdate = require("../../../src/utils/settings-update");
+const settingsUpdate = require("../../../src/helpers/settings-update");
```
timliuhull: Ok, let's sync up on what's going on with settings update.  Because I updated my jest mocks to mock the way when it was in utils.  and looks like we have a couple ways of mocking.  Maybe we can share 1 way

packages/connectors/hull-outreach/server/actions/oauth.js
michaloo: you are right, I wanted to remove them, but after few implementations I decided that's a bad idea, so I changed it here really quick, I hope it will be painless.

PR #3, opened 2018-10-18
I want to see the stack so I know where my error is when i'm debugin

timliuhull:
what do you think?  Was there a reason we weren't showing the stack in debug mode?  I know it isn't always super obvious because of all the async stuff, but it has definitely helped me...

### DIFF
--- a/packages/hull-client/test/integration/retry-test.js
+++ b/packages/hull-client/test/integration/retry-test.js
@@ -33,17 +33,16 @@ describe("client retrying", function test() {
       });
   });
 
-  it("should retry 2 times if get 502 response, then reject", (done) => {
+  it("should retry 2 times if get 502 response, then reject", () => {
     const stub = minihull.stubApp("/api/v1/testing")
       .callsFake((req, res) => {
         res.status(502).end("error 502");
       });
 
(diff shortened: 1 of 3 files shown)


timliuhull:
Unsure why it's failing, it passes tests locally...

michaloo:
Thanks @timliuhull.

About the failing tests -> there is something not deterministic in hull-client tests. Which is interesting since this is the same version as from single repo.

But I will have a look. In one of the recent commits I've made the `testScenario` function much stricter in comparing expected and actual results, so I thought it may be it.

Please let me know about any case of test instability.

michaloo:
Okay, found a solution, but it's super confusing.

When I rerun the tests on CircleCI it fails every time and I couldn't reproduce it on my local machine even running inside the same docker image on limited cpu/memory resources.

When I ssh'ed to the ci instance and tried to debug the problem the tests were failing until I've edited the `lib/rest-api.js` file which is responsible for handling the retry. Even if I just added some random console.log the tests suddenly passed.

It looked like babeljs cache, so I disabled it for testing, but I have no idea how this thing could got into such state since there were no changes to this file recently.

PR #11, opened 2018-11-19
Fix typeform mapping issue

michaloo:


packages/connectors/hull-typeform/manifest.json
```diff
@@ -31,38 +31,17 @@
       }
     },
     {
-      "name" : "incoming_user_attributes",
+      "name": "incoming_user_attributes",
+      "direction": "incoming",
```
timliuhull: What is this? Is this a new attribute?

PR #12, opened 2018-11-21
Release/hull hubspot accounts

michaloo:


michaloo:
@timliuhull ha, I was curious what would you do if I request a review from you :)
but definitely let's do a walkthrough together on a call.

packages/hull/src/handlers/schedule-handler/factory.js
```diff
@@ -80,14 +80,14 @@ function scheduleHandlerFactory(
 
     if (fireAndForget === true) {
       callbackResult.catch(error => {
+        // all TransientErrors (and child error classes such as ConfigurationError)
         if (error instanceof TransientError) {
```
timliuhull: doesn't this need to be (error instanceof TransientError || error instanceof ConfigurationError)

michaloo: let me respond in javascript:
```
class ParentClass {}
class ChildClass extends ParentClass {}
const testInstance = new ChildClass();

(testInstance instanceof ParentClass) === true;
(testInstance instanceof ChildClass) === true;
```

timliuhull: ah, i didn't know ConfigurationError extended TransientError

michaloo: In my understanding it's kind of transient error which can be gone after `ship:update` (aka configuration change), what do you think?

PR #13, opened 2018-11-27
Change Typeform mapping field name

unity:
Change mapping field name to fit with https://github.com/hull/hull-admin/pull/594#discussion_r236605674

### DIFF
--- a/packages/connectors/hull-typeform/manifest.json
+++ b/packages/connectors/hull-typeform/manifest.json
@@ -32,13 +32,13 @@
     },
     {
       "name": "incoming_user_attributes",
-      "direction": "incoming",
       "title" : "List of form fields to save into Hull as user attributes",
       "description": "Note: Date fields should be saved to traits ending with _date suffix",
       "type": "array",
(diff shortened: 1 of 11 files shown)


packages/connectors/hull-typeform/manifest.json
```diff
@@ -36,7 +36,7 @@
       "title" : "List of form fields to save into Hull as user attributes",
       "description": "Note: Date fields should be saved to traits ending with _date suffix",
       "type": "array",
-      "format": "trait-mapping",
+      "format": "traitMapping",
       "options": {
         "allowOverwrite": false,
```
michaloo: we should rename that to `showOverwriteToggle`, right?

unity: you're right. Updated

michaloo:
@unity let's do:
`bash scripts/version-bump.sh 0.0.7-monorepo` to perform a global version bump including manifest.jsons so we need to know if the connector manifest.json was migrated or not.

PR #19, opened 2018-12-07
Cleaner HubSpot docs + updated description

edfryed:
From this Paper doc here: https://paper.dropbox.com/doc/Connector-Doc-Example-Two-Way-Sync-Connector-HubSpot--ATMOuwtH4PB_VNEgv2xzyq1yAg-HM3E8hyxDObb3RkSwWoGB

Standardised the docs + removed typos/spelling + improved readability.

Also updated the connector description.


packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
+2. If so, updates the existing Hull User
+3. Else, creates a new Hull User
 
-- **fetch accounts** - defined if connector will try to read and fetch any `Hubspot Company` information
-- **incoming account identity** - specify which `Hubspot Company Properties` we will use to identify `Hull Account`
-- **incoming account attributes** - defines which `Hubspot Company Properties` are stored in `Hull Account Attributes`
-- **incoming user linking** - defines if Hull will link `Hull User` with `Hull Account` depending on `Hubspot Contact` to `Hubspot Company` association
+### Fetching HubSpot Companies into Hull
 
+HubSpot Companies are not handled by default. This must be enabled in the Connector Settings under the **Fetch accounts?** toggle.
```
timliuhull: Remove question mark?

edfryed: Matches the content in the connector. Agree we should probably standardise this (without the question mark) in the connector settings and the docs.

Will remove

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
-### Outgoing Accounts
+HubSpot Contacts and their updates are fetched into Hull manually or on a schedule.
 
-- **outgoing account segments** - defines which account segments are sent to Hubspot
-- **outgoing account attributes** - defines which `Hull Account Attributes` are updated/created on `Hubspot Company`
-- **outgoing user linking** - defines if Hull will associate `Hubspot Contact` to `Hubspot Company` (see below how the linking is performed)
+When new data about a HubSpot Contact is [ingested](https://www.hull.io/docs/data_lifecycle/ingest/), Hull follows this process:
 
-### Incoming Accounts
+1. Checks if there is a matching `anonymous_id` (HubSpot Contact `VID`) or `email`
```
michaloo: This behavior depends on incoming user identifier configuration, it's `email` by default.

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
 
-|Hull Entity|Hubspot Entity|
-|-----------|--------------|
-|User       |Contact       |
-|Account    |Company       |
+When new data about a HubSpot Company is ingested, Hull follows this process:
 
-No other objects besides the ones listed above are supported. If you need to synchronize additional objects please reach out to our customer success team to explore the options on a case-by-case basis.
+1. Check if the required identifier is present. Skip if not.
+2. Checks if there is a matching `domain` in Hull
```
michaloo: This depends on incoming accounts identifiers configuration, domain is default.

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
 
-- **outgoing account segments** - defines which account segments are sent to Hubspot
-- **outgoing account attributes** - defines which `Hull Account Attributes` are updated/created on `Hubspot Company`
-- **outgoing user linking** - defines if Hull will associate `Hubspot Contact` to `Hubspot Company` (see below how the linking is performed)
+When new data about a HubSpot Contact is [ingested](https://www.hull.io/docs/data_lifecycle/ingest/), Hull follows this process:
 
-### Incoming Accounts
+1. Checks if there is a matching `anonymous_id` (HubSpot Contact `VID`) or `email`
+2. If so, updates the existing Hull User
+3. Else, creates a new Hull User
```
michaloo: I think it's also important to mention here that a merge can happen.

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
-|-----------|--------------|
-|User       |Contact       |
-|Account    |Company       |
+When new data about a HubSpot Company is ingested, Hull follows this process:
 
-No other objects besides the ones listed above are supported. If you need to synchronize additional objects please reach out to our customer success team to explore the options on a case-by-case basis.
+1. Check if the required identifier is present. Skip if not.
+2. Checks if there is a matching `domain` in Hull
+3. If so, updates the existing Hull Account
+4. Else, creates a new Hull Account
```
michaloo: I think it's also important to mention here that a merge can happen - especially in case of accounts, if customer brings in external_id.

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
 
-Accounts synchronization is disabled by default. To enable it you need to setup **fetch accounts** and **outgoing account segments** settings (look above).
+### Linking HubSpot Contacts & Companies in Hull
 
-## Identity Resolution
-The Hubspot connector uses upsert policies for synchronizing data. If we cannot find a matching record, a new one will be created automatically. This section explains how we match existing records between both systems.
+Hull will match Users and Accounts by a common `domain`.
```
michaloo: What does it mean?

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
+If the HubSpot Company does not have a required identifier, the User will be associated with an empty Hull Account.
+
+### Creating & Updating HubSpot Contacts
+
+Hull Users must be in a whitelisted User Segment to be synced to HubSpot. By default, no Users are synced.
+
+When Hull publishes a [User Update](https://www.hull.io/docs/data_lifecycle/notify/#format-of-a-user-update-notification) for a User whitelisted to sync to HubSpot:
+
+1. Check if there is an `email`. Skip if not.
+2. Check if there is a matching `anonymous_id` (corresponds to a HubSpot Contact `VID`) or `email` in HubSpot
```
michaloo: Precisely speaking we use the `hubspot/id` trait for the outgoing traffic matching.

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
+3. If so, update existing HubSpot Contact
+4. Else, create new HubSpot Contact
```
michaloo: I think it's important to mention that this behavior is handled on Hubspot end, it's not a connector logic, but Hubspot API endpoint: https://developers.hubspot.com/docs/methods/contacts/batch_create_or_update

> Each set of property data must include either an email address or a vid. When the vid is used, the existing contact record with that vid will be updated. When an email address is used, and there is an existing contact with the same email address, that contact will be updated. If an email address is used, and there is not an existing record with that email address, then a new contact record will be created.
If you're working with a known contact, using the vid is preferred.

packages/connectors/hull-hubspot/assets/readme.md
```diff
@@ -1,140 +1,210 @@
...
+
+### Creating & Updating HubSpot Companies
+
+Hull Accounts must be in a whitelisted Account Segment, or be associated with a whitelisted User to be synced to HubSpot. By default, no Accounts are synced.
```
michaloo: The connector will only create/update Company object if the account matches the segment, the association with whitelisted user is not affecting the filter here.

Maybe this point was about being able to map account traits into Hubspot contact properties?

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Removed.

How is the HubSpot Contact VID handled? (Anonymous ID?), if at all?

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Will modify to keep it consistent with the explanation above.

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Not sure how best to explain the scenarios where a merge can occur - it's an edge case?

packages/connectors/hull-hubspot/assets/readme.md
michaloo: @edfryed sorry, I wasn't too precise here. Indeed `HubSpot Contact VID` is always stored as `Anonymous ID` no matter what's the customer configuration. I was only referring to the `email`.

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Which matches to HubSpot Contact VID?

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Make sense. This was originally written like the ingestion flow where Hull handles it. Will update.

packages/connectors/hull-hubspot/assets/readme.md
michaloo: yes, when we fetch stuff we store HubSpot Contact VID as both anonymous_id and `hubspot/id` trait. Then when going out we use the trait.

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Okay - will remove the reference

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Clarified

packages/connectors/hull-hubspot/assets/readme.md
edfryed: Updated.

PR #23, opened 2019-01-11
Update filter settings titles

sbellity:
Part of https://github.com/hull/hull-connectors-issues/issues/184


timliuhull:
Looks good!

PR #30, opened 2019-01-25
Feature/new outreach connector development

timliuhull:
Now that we've fixed the testing framework, we can merge to develop


package.json
```diff
@@ -22,7 +22,7 @@
     "dev": "sh scripts/start-dev-connector.sh",
     "test": "yarn test:lint && jest && yarn workspaces run test",
     "test:lint": "eslint packages",
-    "test:ci": "yarn test:lint && jest && nyc --babel-cache --hook-run-in-context false yarn workspaces run test",
+    "test:ci": "yarn test:lint && jest --runInBand && nyc --babel-cache --hook-run-in-context false yarn workspaces run test",
```
timliuhull: I added this so that the tests will run 1 at a time, it helps so that we don't run into memory errors

packages/hull-client/test/integration/retry-test.js
```diff
@@ -84,7 +84,7 @@ describe("client retrying", function test() {
         res.end("ok")
       });
 
-    return client.get("/testing", {}, { timeout: 20, retry: 10 })
+    return client.get("/testing", {}, { timeout: 100, retry: 10 })
```
timliuhull: This was timing out intermittently before the second call could be made, so i upped the timeout a little

packages/hull-connector-framework/src/test-scenario-runner.js
```diff
@@ -156,6 +158,10 @@ class TestScenarioRunner extends EventEmitter {
     this.minihull = new Minihull();
     this.app = express();
     this.rawScenarioDefinition = scenarioDefinition;
+
+    if (debounceWait) {
+      this.debounceWait = debounceWait;
```
timliuhull: I'm not using this anymore, it was too unstable, should we take this out?  Might be a nice feature for the future?  We may need to work out this whole timing concept anyway

packages/hull-connector-framework/src/test-scenario.js
```diff
@@ -8,16 +8,18 @@ function testScenario(
...
     connectorWorker?: Function,
-    connectorManifest: Object
+    connectorManifest: Object,
+    debounceWait?: number
   },
   scenarioDefinition: TestScenarioDefinition
 ): Promise<*> {
   const runner = new TestScenarioRunner(
-    { connectorServer, connectorWorker, connectorManifest },
+    { connectorServer, connectorWorker, connectorManifest, debounceWait },
```
timliuhull: Should we take this out?  Relates to debounce config feature

packages/hull/src/handlers/html-handler/factory.js
```diff
@@ -0,0 +1,151 @@
+// @flow
```
timliuhull: I wasn't sure how we wanted to handle this.  But for the most part this is just a copy of jsonHandler...

packages/hull/src/handlers/incoming-request-handler/factory.js
```diff
@@ -12,6 +12,7 @@ type HullIncomingRequestHandlerCallback = (
 ) => Promise<*>;
 type HullIncomingRequestHandlerOptions = {
   disableErrorHandling?: boolean,
+  parseCredentialsFromQuery?: boolean,
```
timliuhull: Added this option in cases where webhooks contain the credentials

packages/hull/src/handlers/json-handler/factory.js
```diff
@@ -88,6 +91,23 @@ function jsonHandlerFactory(
...
+            `error_message:${_.snakeCase(error.message)}`
+          ]);
+        } else {
+          req.hull.metric.captureException(error);
+        }
+      });
+      res.json({ status: "deferred" });
+      return;
+    }
+
```
timliuhull: I think we talked through this logic and were ok with it.  But want to take one last look

packages/hull/src/helpers/settings-update.js
```diff
@@ -33,7 +33,7 @@ function settingsUpdate(
     if (!cache) {
       return connector;
     }
-    return cache.del(connector.id).then(() => connector);
+    return cache.del("connector").then(() => connector);
```
timliuhull: This was DEFINITELY a bug where on settings up date in certain scenarios we weren't removing the connector settings from the cache because we need to do it on the "connector" string.  But should take another look.  This was an early change that I guess didn't get committed to the master monorepo

yarn.lock
```diff
@@ -12619,7 +12624,7 @@ uri-js@^4.2.2:
   dependencies:
     punycode "^2.1.0"
 
-urijs@^1.19.1:
+urijs@^1.18.0, urijs@^1.19.1:
```
timliuhull: These updates to yarn.lock seem correct.  Right?

packages/hull-connector-framework/src/test-scenario-runner.js
michaloo: I would say we can leave it. Generally the value of `debounceWait` should be smallest possible but greater than the longest time between two external (side effect) operations of the connector (doing a firehose call, platform api call, nock call or loggin stuff),

packages/hull/src/handlers/html-handler/factory.js
```diff
@@ -0,0 +1,151 @@
...
+  HullJsonHandlerCallback,
+  HullJsonHandlerOptions
+>;
+
+/**
+ * TODO the logic for this should be combined with jsonHandler
+ * should we use inheritance, composition or a functional approach to keep this code DRY?
+ * or should we abide by the abstraction rule of 3, and allow this second implementation to be ok for now
+ *
+ * This handler allows to handle simple, authorized HTTP calls.
```
michaloo: and return a rendered HTML document, right?

packages/hull/src/helpers/settings-update.js
michaloo: if we have `cache.set("connector")` in the other middleware then we are good here.

packages/hull/src/handlers/html-handler/factory.js
michaloo: I asked one thing from the inline docs below.

PR #31, opened 2019-01-30
updated Hubspot status endpoint with more descriptive messages and us…

timliuhull:
…ing warning more


packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,43 +22,75 @@ function getMessageFromError(err): string {
...
-    pushMessage("Missing API token.");
+    pushMessage("error", 'No OAuth AccessToken found.  Please make sure to allow Hull to access your Hubspot data by clicking the "Credentials & Actions" button on the connector page and following the workflow provided');
   }
 
-  if (!_.get(connector, "private_settings.refresh_token")) {
-    pushMessage("Missing refresh token.");
-  }
+  // This doesn't really matter either.
+  // If there's a real problem, we'll hit Unauthorized when doing the below tests
+  // if (!_.get(connector, "private_settings.refresh_token")) {
```
michaloo: There is a theoretical situation that we fail to get `refresh_token` or `portal_id` during oAuth, but we got the token, I think in this case we could say error and direct the user into reauthorizing the connector (there is a button on the Credentials & Actions screen)

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -82,34 +114,51 @@ function statusCheckAction(ctx: HullContext) {
         .then(body => {
           if (!_.find(body, g => g.name === "hull")) {
             pushMessage(
-              "Hubspot is not properly configured. Missing hull group"
+              "warning",
+              "Hubspot is missing the custom attribute for Hull groups.  Initial synch with Hubspot may not have been completed yet"
```
michaloo: Maybe it can read: `Hubspot is missing group of properties to store Hull attributes`?

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,43 +22,75 @@ function getMessageFromError(err): string {
...
+  //   pushMessage("Missing portal id.");
+  // }
 
-  if (!_.get(connector, "private_settings.portal_id")) {
-    pushMessage("Missing portal id.");
+  if (
+    _.isEmpty(_.get(connector, "private_settings.synchronized_user_segments", []))
+    || _.isEmpty(_.get(connector, "private_settings.synchronized_account_segments", []))
+  ) {
+    pushMessage("warning",
```
michaloo: Do you think we could make this `ok` status? Since it's perfectly valid scenario to use only incoming flow on this connector.

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: Does it matter if we still have the token?  I guess we won't be able to refresh.  How much of an edge case is this? How often does this happen?  Do we know why it happens?

Depending on the answer to those questions we can come up with a good message

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: Ah... so we store all of the Hull specific attributes in this special group?

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: Sure, I was just going with what we normally do. I think this is fine for now, at least it's more consistent.  Either way, I'm fine doing they way you like.

packages/connectors/hull-hubspot/server/actions/status-check.js
michaloo: If we don't have refresh_token the connector will become non-operational in few hours after the token expires and we cannot refresh it.
I think it's very rare case, maybe it ever happened once to me that something broke down during oauth flow and we got partial response from hubspot, but I assumed that since refresh_token is critical for the connector we can double check it here.

packages/connectors/hull-hubspot/server/actions/status-check.js
michaloo: yes, in outgoing attributes mapper you can either pick existing field which can be in any group, or you can create new custom field which will be only added to Hull group, nowhere else.

packages/connectors/hull-hubspot/server/actions/status-check.js
michaloo: then I would like to review them last time with Sven to get his approval. We could produce a table `| status | message |`, but Sven is probably okay with looking at PR.

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: fixed

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: Updated with new wording.

timliuhull:
@SMK1085 Your brand new hubspot messaging.  I'm unsure how to describe some of the problems in details that the customer understands, but most of the messaging has possible resolution.  even if it means "contact support" which I tried to only use as last case.

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,58 +22,115 @@ function getMessageFromError(err): string {
...
+    _.isEmpty(
+      _.get(connector, "private_settings.synchronized_user_segments", [])
+    ) ||
+    _.isEmpty(
+      _.get(connector, "private_settings.synchronized_account_segments", [])
+    )
+  ) {
+    pushMessage(
+      "warning",
+      "No users or accounts will be sent from Hull to Hubspot because there are no whitelisted segments configured.  Please visit the connector settings page and add segments to be sent to Hubspot"
```
SMK1085: Doesn't really make sense, this is not a warning at all! Please get rid of that one. As @michaloo already mentioned it is perfectly valid to have incoming data flows. It doesn't need to be consistent with the stuff we have (which is imo completely backwards and no customer understands it), it needs to be correct and reliable. A warning should only be emitted when something goes wrong intermittently. So in consequence, the status endpoint could only emit a warning when there is a connection error with the Hubspot API.

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,58 +22,115 @@ function getMessageFromError(err): string {
...
+    );
   }
 
   if (
-    _.isEmpty(_.get(connector, "private_settings.sync_fields_to_hubspot", []))
+    _.isEmpty(
+      _.get(connector, "private_settings.outgoing_user_attributes", [])
```
SMK1085: This doesn't make sense as a standalone validation. This should only be an error if I have segments configured, if not, it's perfectly fine to have no outgoing attributes configured.

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,58 +22,115 @@ function getMessageFromError(err): string {
...
+      _.get(connector, "private_settings.incoming_user_attributes", [])
+    ) ||
+    _.isEmpty(
+      _.get(connector, "private_settings.incoming_account_attributes", [])
+    )
+  ) {
     pushMessage(
-      "No fields are going to be sent from hubspot to hull because of missing configuration."
+      "warning",
+      "There are no attributes configured to be pulled into Hull.  Visit the connector settings page and configure the attributes for Accounts/Users that you want imported"
```
SMK1085: The sentence `There are no attributes configured to be pulled into Hull` is incorrect, we pull default fields as always. This should not be a warning at all, because it implies that everyone has custom fields that they want to sync to Hull which is not true.

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,58 +22,115 @@ function getMessageFromError(err): string {
...
     );
   }
 
-  if (_.isEmpty(_.get(connector, "private_settings.sync_fields_to_hull", []))) {
+  if (
+    _.isEmpty(
+      _.get(connector, "private_settings.incoming_user_attributes", [])
+    ) ||
+    _.isEmpty(
+      _.get(connector, "private_settings.incoming_account_attributes", [])
```
SMK1085: We cannot do this in one check, you can disable the fetch of accounts. Hence this needs to be split and this should only be an informational message anyways not a warning. See my comment on the message below. I am sorry, but how would you feel if your connector shows `Warning` all the time even you have configured everything properly and just want to run with the default fields?

packages/connectors/hull-hubspot/server/actions/status-check.js
```diff
@@ -22,58 +22,115 @@ function getMessageFromError(err): string {
...
   if (_.get(connector, "private_settings.token")) {
     promises.push(
       syncAgent.hubspotClient
         .getRecentlyUpdatedContacts()
         .then(results => {
           if (results.body.contacts && results.body.contacts.length === 0) {
-            pushMessage("Got Zero results when fetching contacts.");
+            pushMessage(
+              "warning",
+              'The Hubspot organization that this connector is communicating with does not contain contacts.  If you think this is not correct, please try and re-authenticate with the "Credentials/Action" button.  If this warning persists please contact your Hull support representative'
```
SMK1085: This is not correct, the organization might have contacts but none of them has been updated since the last fetch. So why do we need to pollute the timeline of the connector with this? This information is available in the logs, please ditch it entirely!

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: yea, this is shitty because it's yet another thing that depends on which connector we're talking about.  i don't like that there's these magical default fields you can't see.

packages/connectors/hull-hubspot/server/actions/status-check.js
timliuhull: I like that... we don't send default outgoing probably right?

PR #32, opened 2019-02-06
Release/outreach connector to master

timliuhull:
New Outreach connector to master


packages/hull/src/helpers/settings-update.js
```diff
@@ -33,7 +33,7 @@ function settingsUpdate(
     if (!cache) {
       return connector;
     }
-    return cache.del(connector.id).then(() => connector);
+    return cache.del("connector").then(() => connector);
```
timliuhull: Needed to be corrected because we're not using id to cache connector anymore.  Refer to:
https://github.com/hull/hull-connectors/blob/master/packages/hull/src/middlewares/full-context-fetch.js#L18

packages/hull/src/middlewares/full-context-fetch.js
```diff
@@ -24,7 +24,7 @@ function fetchConnector(ctx): Promise<*> {
   );
 }
 
-function fetchSegments(ctx, entityType = "users") {
+function fetchSegments(ctx, entityType = "user") {
```
timliuhull: Seems right based off how it's called in other places.  But can't find a place where it's called without entityType.

Here
https://github.com/hull/hull-connectors/blob/master/packages/hull/src/middlewares/full-context-fetch.js#L84

packages/hull/src/handlers/json-handler/factory.js
```diff
@@ -12,7 +12,8 @@ type HullJsonHandlerOptions = {
     options?: Object
   },
   disableErrorHandling?: boolean,
-  respondWithError?: boolean
+  respondWithError?: boolean,
+  fireAndForget?: boolean
```
timliuhull: Added fire and forget for jsonhandler for cases when we hit fetchAll using jsonHandler

packages/hull/src/handlers/incoming-request-handler/factory.js
```diff
@@ -12,6 +12,7 @@ type HullIncomingRequestHandlerCallback = (
 ) => Promise<*>;
 type HullIncomingRequestHandlerOptions = {
   disableErrorHandling?: boolean,
+  parseCredentialsFromQuery?: boolean,
```
timliuhull: credentials from query for webhooks!

michaloo: which generally may be something we would like to avoid - to send out our secrets and ids out to some external services.

PR #37, opened 2019-02-17
Hotfix/tab names default status

unity:


unity:
@timliuhull please review & merge - it's only frontend code and brings real improvements to the ui consistency

packages/connectors/hull-hubspot/manifest.json
```diff
@@ -104,44 +106,46 @@
...
+      "title": "User Outgoing Setup",
+      "description": "How we send Users as Hubspot Contacts",
       "format": "title",
       "type": "string"
     },
     {
-      "name": "synchronized_account_segments",
-      "title": "Account Filter",
-      "description": "Defines the list of Account Segments to send. Nothing sent by default.",
+      "name": "synchronized_user_segments",
```
timliuhull: this looks a little weird.  This is only because you moved it to the bottom right?

packages/connectors/hull-mailchimp/manifest.json
```diff
@@ -6,7 +6,15 @@
...
   "logo": "logo.png",
   "picture": "picture.png",
-  "admin" : "/auth/",
+  "tabs": [
+    {
+      "title": "Credentials & Actions",
+      "url": "/auth/",
+      "size": "small",
+      "setup": true,
+      "editable": false
```
timliuhull: This is a user driven behavior and not part of the data flow correct?  So even if this broke something, we could fix it quickly

timliuhull:
Would be cool if we could easily have something that could allow commits by a single person into ui directories or something

unity:
Not sure I understand the above comme t

packages/connectors/hull-mailchimp/manifest.json
unity: Yes exactly. Only impacts dashboard usage

packages/connectors/hull-hubspot/manifest.json
unity: Yes. I reorganized the settings to be clearer

timliuhull:
> Not sure I understand the above comment

nevermind, just thinking about how to institute faster pushes for ui tweaks

packages/connectors/hull-customerio/manifest.json
```diff
@@ -37,18 +37,11 @@
     {
       "name": "synchronized_segments",
```
timliuhull: We need to do a version bump on the repo

PR #40, opened 2019-03-05
new manifest options

timliuhull:


packages/connectors/hull-outreach/manifest.json
```diff
@@ -5,52 +5,58 @@
...
-
   "tags": [
     "incoming",
     "outgoing",
     "batch",
     "batch-accounts",
     "oneColumn",
-    "smart-notifier"
+    "smart-notifier",
+    "betterforms"
```
timliuhull: This made the most sense in keeping with the "tags" convention to add the flag to use the new form flag

packages/connectors/hull-outreach/manifest.json
```diff
@@ -5,52 +5,58 @@
...
   ],
 
   "private_settings": [
     {
       "name": "webhook_id",
       "title": "Outreach webhook id",
       "type": "string",
-      "format": "hidden"
+      "format": "hidden",
+      "tags": ["authentication", "webhook_id"]
```
timliuhull: I don't think it matters for the UI if we tag this kind of stuff, but I think it may mean something in the future to determine the type of validation we can do on particular fields

packages/connectors/hull-outreach/manifest.json
```diff
@@ -5,52 +5,58 @@
...
     },
     {
       "name": "access_token",
       "title": "OAuth Access Token",
       "type": "string",
-      "format": "hidden"
+      "format": "hidden",
+      "tags": ["authentication", "access_token"]
```
timliuhull: This one in particular, could be interesting to infer certain things about the state of the connector if it isn't set across all connectors.  Maybe we could change it to a generic "api_token" which could be used to validate both oauth and api key implementations

packages/connectors/hull-outreach/manifest.json
```diff
@@ -5,52 +5,58 @@
...
-      "format": "hidden"
+      "format": "hidden",
+      "tags": ["authentication"]
     },
     {
       "name": "user_claims",
       "title": "User Identity",
       "type": "array",
       "format": "table",
+      "tags": ["incoming", "outgoing", "identity", "user"],
```
timliuhull: This claims table applies to incoming and outgoing.  I added "identity" and "user" as sub tags that we could use for interesting stuff in the future

packages/connectors/hull-outreach/manifest.json
```diff
@@ -80,19 +86,22 @@
       "name": "user_claim_hero",
       "type": "string",
       "headerTemplate": "Mapped user identity attributes will automatically be pulled in the incoming and outgoing traffic",
-      "format": "information"
+      "format": "information",
+      "tags": ["incoming", "outgoing", "identity", "user"]
```
timliuhull: I'm tagging the information fields too.  Not sure how useful or easy to use this would be for identifying interesting description fields.  I'd probably just rather use the embedded description in each component.  The only reason this was added for the "user_claims" component is because it's of format=table, type=array, which it seems does not support "description"

packages/connectors/hull-outreach/manifest.json
```diff
@@ -249,13 +271,15 @@
       "url": "/auth",
       "size": "small",
       "editable": false,
-      "setup": true
+      "setup": true,
+      "tags": ["authenticaiton"]
```
timliuhull: I added a "tags" to the tabs property too.  Won't need it for this iteration, but will probably need it for the wizard in the future

packages/connectors/hull-outreach/manifest.json
unity: Under no circumstances should the same setting be in multiple places. That’s a huge issue from a UX standpoint. If using tags, and the manifest model doesn’t hard-enforce this we need to enforce it somehow.

packages/connectors/hull-outreach/manifest.json
unity: I have added descriptions support to the array field in my “validation” branch of the dashboard.

packages/connectors/hull-outreach/manifest.json
```diff
@@ -249,13 +271,15 @@
...
+      "setup": true,
+      "tags": ["authentication"]
     },
     {
       "title": "Actions",
       "url": "/admin",
       "size": "small",
-      "editable": false
+      "editable": false,
+      "tags": ["fetch"]
```
unity: Where is the spec that describes what tags mean and what is their impact ?
Also I would very much prefer if we exposed actions as buttons Instead of html pages with buttons inside. Seems like a simple thing to do but would radically simplify the UX. For this we could expose 
```
actions: [{ name: “Fetch Users”, url: “/fetch-user”, method: “POST”, prerequisite: { api_key: { pattern: “\d+” } }]
```

unity: Just came up with a format on the fly but I would like this to be soecced and validated for consistency before we implement

packages/connectors/hull-outreach/manifest.json
timliuhull: What do you mean "same setting be in multiple places"?  Do you mean because this has the incoming and outgoing tag?  It's because this user claims field applies to both incoming and outgoing data...

packages/connectors/hull-outreach/manifest.json
timliuhull: yes, we need that, but I was thinking that would be the next interation.  First the tags for grouping so that Michael can key off of it to do the Grouped settings work, then expose the actions that we'll need for the wizard when we start work on that

packages/connectors/hull-outreach/manifest.json
timliuhull: cool!

PR #44, opened 2019-03-12
updated filter logic to not send account linking if the user is not i…

timliuhull:
…n segment

### DIFF
--- a/packages/connectors/hull-outreach/server/shared/utils.js
+++ b/packages/connectors/hull-outreach/server/shared/utils.js
@@ -77,25 +77,6 @@ function toSendMessage(context: HullContext, targetEntity: "user" | "account",
       throw new Error(`Invalid input for target entity, option not supported: ${targetEntity}`);
     }
 
-    // Should we do on entered segment too?
-    if (targetEntity === "user") {
-
-      const linkInService = _.get(context, "connector.private_settings.link_users_in_service");
-
-      if (!isUndefinedOrNull(linkInService) && typeof linkInService === "boolean" && linkInService) {
-        const changedAccounts = _.get(message, `changes.account.id`);
-
-        if (!_.isEmpty(changedAccounts)) {
-          // user changed accounts, and link_users_in_service is enabled
-          // so update user...
-          return true;
-        }
-        // if account enteres a synchronized segment
-        // but it was or wasn't in a synchronized segment before
-        // may want to perform account linking
-      }
-    }
-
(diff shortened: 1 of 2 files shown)


packages/connectors/hull-outreach/server/shared/utils.js
```diff
@@ -77,25 +77,6 @@ function toSendMessage(context: HullContext, targetEntity: "user" | "account",
...
-          // user changed accounts, and link_users_in_service is enabled
-          // so update user...
-          return true;
-        }
-        // if account enteres a synchronized segment
-        // but it was or wasn't in a synchronized segment before
-        // may want to perform account linking
-      }
-    }
-
```
timliuhull: This logic should be after checking that the user is in a synchronized segment, otherwise if we have "link_in_service" and there looks like there's an account change (when changes.account.id is not empty), then we would send the user, even if it isn't in a segment

packages/connectors/hull-outreach/server/shared/utils.js
```diff
@@ -145,6 +126,25 @@ function toSendMessage(context: HullContext, targetEntity: "user" | "account",
       return false;
     }
 
+    // Should we do on entered segment too?
```
timliuhull: Moving this logic after segment filter logic

packages/connectors/hull-outreach/test/integration/scenarios/outgoing-user-link-new-account-filter.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+test("should not link account in outreach if user is not in segment", () => {
+  return testScenario({ connectorServer }, ({ handlers, nock, expect }) => {
+    const updateMessages = _.cloneDeep(require("../fixtures/notifier-payloads/outgoing-user-link-new-account.json"));
+    updateMessages.connector.private_settings.synchronized_user_segments = [];
+    _.set(updateMessages.messages[0], "changes.account.id", [
+                        null,
+                        "someid"
+                    ]);
```
timliuhull: This test uses the same notifier-payload as other tests, but in this case we modify it so that there are no synchronized_user_segments, and we make sure that an account change is detected

timliuhull: Maybe we should put this in a comment on the code so we remember next time what its for

michaloo:
To make sure I understood the change: the connector was checking if the account was changed and whether it should relink the user in 3rd party and it was ignoring the user segment filter, correct?

PR #55, opened 2019-03-21
add default fields to readme

andyh565:


michaloo:
@andyh565 nice, can we make it a table instead of this list?

michaloo:
@andyh565 @timliuhull I think we can merge it to develop instead and prepare a release including couple of changes for hubspot:
1. full removal of overwrite flag (including code and documentation)
2. documentation update with the default mapping

Also @andyh565 we need to document the companies default mapping too, is it already included?

andyh565:
> @andyh565 nice, can we make it a table instead of this list?

@timliuhull 

added tables including both contact and company mappings

PR #58, opened 2019-03-22
update manifests for "first-install"

g10:
https://github.com/hull/hull-admin/issues/701


timliuhull:
Looks fine to push into prod, but lets push into develop first

PR #80, opened 2019-04-24
check zero value for owner

andyh565:


timliuhull:
Backmerging to develop, and adding new transform condition capabilities

PR #119, opened 2019-06-28
Feature/hubspot deals

andyh565:


andyh565:
deals api
https://developers.hubspot.com/docs/methods/deals/deals_overview

PR #120, opened 2019-06-29
Spike/merge

unity:


timliuhull:
Linting complete.  Redis and Mongo addition complete. Only remaining issue is the hull tests, but will continue to merge before test are fixed.  Will continue to fix on develop branch, and will then make the call to suppress tests if needed

PR #143, opened 2019-07-17
updated the manifest to make a little more user friendly

timliuhull:


timliuhull:
yay! more descriptive manifests!

PR #180, opened 2019-08-02
Outreach Error: Not Found

louisjahn:
Fixing this related issue:
https://sentry.io/organizations/hull-dev/issues/1115764434/?project=1306515&query=is%3Aunresolved&statsPeriod=14d

### DIFF
--- a/packages/hull/src/middlewares/full-context-fetch.js
+++ b/packages/hull/src/middlewares/full-context-fetch.js
@@ -116,7 +116,7 @@ function fullContextFetchMiddlewareFactory({
     } catch (error) {
       if (error.status === 404) {
         debug("Connector not found");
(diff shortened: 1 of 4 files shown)


packages/hull/src/middlewares/full-context-fetch.js
```diff
@@ -113,6 +114,9 @@ function fullContextFetchMiddlewareFactory({
       });
       return next();
     } catch (error) {
+      if (error.status === 404) {
+        return next(new ConnectorNotFoundError("Invalid id / secret"));
```
timliuhull: Add debug statement here and comment out this real quick.  Then we can see where it's happening if it'sall ok to supress.  then uncomment the error

PR #184, opened 2019-08-12
Hotfix/mailchimp selectlist step

louisjahn:


timliuhull:
The mailchimp connector is broken right now becase we don't change the status past setupRequired if the list isn't selected, but you can't select the list during setupRequired.  Doing this quick fix to fix it.  But better fix is to have a 2nd step where you have to enable the list.  will check with the dashboard team on how to do this.  but fixing this like this for now

PR #188, opened 2019-08-19
Feature/processor

unity:


packages/hull-client/src/utils/traits.js
```diff
@@ -56,26 +52,27 @@ type HullEntityNested = {
...
+    entity,
+    (grouped, value, key: string) => {
       let dest = key;
       if (key.match(/^traits_/)) {
-        if (key.match(/\//)) {
-          dest = key.replace(/^traits_/, "");
-        } else {
-          dest = key.replace(/^traits_/, "traits/");
-        }
+        dest = key.replace(/^traits_/, "");
```
timliuhull: I guess previous logic was trying to do something that we don't care about anymore?

packages/hull/src/connector/hull-connector.js
```diff
@@ -248,17 +258,19 @@ class HullConnector {
...
+      typeof this.handlers === "function"
+        ? await this.handlers(this)
+        : this.handlers;
     return this._handlers;
   }
 
-  setupRoutes(app: $Application) {
-    const handlers = this.getHandlers();
+  async setupRoutes(app: $Application) {
+    const handlers = await this.getHandlers();
```
timliuhull: why do we need an async get handler? Would we be pulling them from somewhere remote?

packages/hull/src/handlers/external-handler.js
```diff
@@ -78,7 +78,7 @@ const handlerFactory = ({
     }
 
     // There'a an actual Response to send
-    const { pageLocation, data, status = 200, text } = response;
+    const { pageLocation, data, error, status = 200, text } = response;
```
timliuhull: where are we using this ability to return an error in the response?  Not sure I like delegating more responsibility to the underlying code to catch errors, but would need to see the case.  Might be an ok feature, but means there's more catching in underlying code which sometimes can lead to suppressing errors

packages/hull/src/handlers/json-handler/factory.js
```diff
@@ -44,6 +44,7 @@ function jsonHandlerFactory(
     options: {
       credentialsFromQuery: true,
       credentialsFromNotification: false,
+      // cacheContextFetch: false,
```
timliuhull: why did we remove this?  I ask because if we were not fetching the context from the cache before, but we are now, it could make us more vulnerable to dirty cache bugs.  though I'm not opposed to trying to find those bugs anyway

packages/hull/src/utils/get-json-body.js
```diff
@@ -1,4 +1,4 @@
 // @flow
-export default function get(body: mixed): Object {
-  return body !== null && typeof body === "object" ? body : {};
+export default function get(body: mixed): mixed {
```
timliuhull: Why do we not have to do all of the null checks anymore?  could the thing that's calling this be expecting the null check help?

packages/hull/src/utils/get-schemas.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+// import _ from "lodash";
+// import fp from "lodash/fp";
+import type { HullContext } from "../types/index";
+// import { formatEvent } from "./format-event";
+
+const EVENTS_ROUTE = "/search/event/bootstrap";
+const USERS_ROUTE = "/users/schema";
+const ACCOUNTS_ROUTE = "/accounts/schema";
+const get = (route: string) => (ctx: HullContext) => () =>
```
timliuhull: Maybe it's a personal preference, but a factory of a factory of for a function is sort of confusing

timliuhull: maybe i'm just understanding it wrong though...

packages/hull-client/src/utils/traits.js
unity: Sven has been warned. It's a breaking change: 

The attributes in the `traits` group are now at top level. Previously, the processor (and only the processor) was exposing "traits_foo" as "traits.foo" and "traits_foo/bar" as "foo.bar" -> Accounts were exempt from this and had "foo" at top level.

This method was only used in processor, and this change makes it consistent with all the other connectors since we've had an util stripping `traits_` from names for a long time now.

It makes Users and Accounts have the same data structure. But the caveat is that code from the processor that used `traits_` attributes or accessed the `traits` object in the User profile needs to be updated.

I checked all the installs of the nextgen processor, and their code: None of the already installed ones are using it.

It does mean we can't just "swap in the processor" in case the customer's code is trying to access "traits.*" -> but I think this is a good change because it removes the last user-facing traces of the "traits" prefix in the monorepo.

Which means this gives us a migration path to not have them at all on users anymore.

unity: so in resume:
previous processor:
```
user: { traits: { foo: "bar" } }
```

nextgen one:
```
user: { foo: "bar }
```

packages/hull/src/connector/hull-connector.js
unity: We have some connectors that benefit from being able to perform async tasks (mainly, those who are using external dependencies) - in this case, the Mongo models require an async initialization, and we want the models to be ready when initializing the handlers.

packages/hull/src/handlers/external-handler.js
unity: Checkout NEW_README.md, usage is explained. It's a way to easily return an error with a predictable pattern: `status: 500, { error: xxx }`

If we ever decide to make them different, we won't have to change the return data in each connector We can just update the handler middlewares

packages/hull/src/handlers/json-handler/factory.js
unity: At some point, it must have struck me that disabling Cache from JSON was suboptimal. Because for instance in a settings page with 10 select boxes, we'd be firing 10 API calls to the connector which in turn would fire 30 calls to the platform (connector+usersettings+accountsettings) - I wanted to try to rely on the cache here. Your call :)

packages/hull/src/utils/get-json-body.js
unity: It turns out that this was breaking requests that sent the body as a string object.
The Incoming handler can receive plain text and parse it as a CSV for instance. 

When realizing this, It came to me that it was probably a bad idea to assume that all data was JSON objects or null (null being a valid JSON value) and that it should be the connector's responsability to parse the body.

The `bodyParser` already handle this properly when they're applied to the payloads (i.e. JSON handlers and Kraken handlers) - which means we're left with checking if a value is `null` or a JSON object (or array) in the connector

With the "fail early" mindset (https://www.martinfowler.com/ieeeSoftware/failFast.pdf),  attempting to parse a non-object in a connector SHOULD throw an error, which is then catched by the error handler, giving us a uniform way to surface invalid payloads. I don't believe that invalid data should be handled by each connector individually - handling the error should be opt-in, not fail-safe. I.e. without any checks, we have a proper error path that returns structured messages, and we can (if we have to) decide to add additional checks in the code to have a different behaviour.

packages/hull/src/utils/get-schemas.js
unity: https://en.wikipedia.org/wiki/Currying ;) - if it doesn't read well, we might want to update. But i personally grew to like this pattern. Did i go too far here?

unity:
> Does not seem like an insignificant amount of changes. We've tested a deployment of this right? Is the account processor running off this branch?

It is true that it's a rather large PR. However, few changes to the existing codebase. Most of the files are additions, and many are json, png, css files et. Al.

The account processor is running off this branch and in general, it is more of an improvement than a refactor branch for existing services.

Happy to answer more questions to clarify things

PR #191, opened 2019-08-21
Hotfix/hull repl update

michaloo:
Now this parameter is required.


yarn.lock
```diff
@@ -7108,6 +7108,11 @@ invariant@^2.2.2, invariant@^2.2.4:
   dependencies:
     loose-envify "^1.0.0"
 
+invert-kv@^1.0.0:
```
timliuhull: @michaloo Does this change require this new library?

PR #218, opened 2019-09-03
Delete old user and reassign user events when detecting a merge event

louisjahn:


packages/connectors/hull-warehouse/server/postgres-sequalize-service.js
```diff
@@ -380,6 +384,52 @@ class SequalizeSdk {
...
+      .findAll({
+        where: {
+          user_id: merged_id
+        },
+        attributes: [
+          "user_id"
+        ]
+      })
+      .then((rows: Array) => {
+        return rows.length === 0 ? [] : _.rows.map(rows, event => {
```
louisjahn: Reassign user events to the destination user id

PR #234, opened 2019-09-05
Feature/hubspot entity deletion

andyh565:


packages/connectors/hull-hubspot/server/actions/check-cache.js
```diff
@@ -0,0 +1,29 @@
+/* @flow */
+import type { HullContext, HullExternalResponse } from "hull";
+
+const _ = require("lodash");
+
+/**
+ * Checks cache and resets value
+ */
+async function checkCache(ctx: HullContext): HullExternalResponse {
```
andyh565: TODO: cache is a list of credentials

packages/hull/src/middlewares/credentials-from-query.js
```diff
@@ -61,6 +63,27 @@ function generateEncryptedToken(clientCredentials, secret) {
   return encrypt(clientCredentials, secret);
 }
 
+async function getClientCredentials(clientCredentials, req) {
```
andyh565: TODO remove

packages/hull/test/unit/crypto.spec.js
```diff
@@ -37,8 +37,11 @@ describe("crypto", () => {
     const res = {};
     const next = sinon.spy();
     const mw = middleware();
```
andyh565: TODO revert once getClientCredentials is removed from middleware

andyh565:
TODO - send response back to Hubspot to avoid retries

PR #243, opened 2019-09-12
added logic to send if we do not have an id for the service, meaning …

timliuhull:
Look at all that beautiful filter logic in 1 place, with unit tests.  In production only logging for hubspot

### DIFF
--- a/packages/hull-connector-framework/src/purplefusion/utils.js
+++ b/packages/hull-connector-framework/src/purplefusion/utils.js
@@ -357,10 +357,23 @@ function toSendMessage(
   // or could be because it's a new connector which we haven't done a full fetch
   const serviceName = _.get(options, "serviceName");
   if (serviceName) {
-    console.log("servicename");
     const serviceId = _.get(message, `${targetEntity}.${serviceName}/id`);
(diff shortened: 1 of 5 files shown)


timliuhull:
Added a log temporarily to find this scenario.  otherwise, won't be able to confirm it well in prod

packages/hull-connector-framework/src/purplefusion/utils.js
```diff
@@ -349,6 +351,33 @@ function toSendMessage(
     return false;
   }
 
+  // in cases where we do not have a id of the service, which means we haven't sync'd the entity before
+  // send the user because we know it's in the list to send
+  // this may be the result of pushing a full segment after it's creation
+  // or could be because it's a new connector which we haven't done a full fetch
+  const serviceName = _.get(options, "serviceName");
+  if (serviceName) {
+    const serviceId = _.get(message, `${targetEntity}.${serviceName}/id`);
```
unity: I think it could be a better practice to rely on the anonymous ID for this service (`${serviceName}:id`”). What do you think ?

PR #245, opened 2019-09-15
logging

andyh565:


packages/connectors/hull-slack/server/handlers/user-update.js
```diff
@@ -133,6 +136,27 @@ const update = (connectSlack: ConnectSlackFunction) => async (
                     entity: ENTITY,
                     attachements
                   });
+                  try {
+                    const userId = _.get(message, "user.id", null);
```
unity: Cosmetic suggestions:
```
const { id } = message.user
if (!id) {...

...
const triggers = messages.map(m => _.pick(m, [“changes”, “events”]));
ctx.client.logger.info(“Duplicate Notification”, { key, triggers });
```

andyh565: thanks, I'll update it

unity: Nothing’s broken, just sharing the syntax I usually use ;) - don’t feel like you have to !

PR #249, opened 2019-09-17
Feature/pipedrive oauth get users

andyh565:


packages/connectors/hull-pipedrive/server/glue.js
```diff
@@ -0,0 +1,264 @@
...
+    ])
+  ],
+  webhooks: ifL(input("body"), route("handleWebhook", cast(WebPayload, input("body")))),
+  handleWebhook:
+    ifL(cond("isEqual", "organization", input("meta.object")), {
+      do: hull("asAccount", input()),
+      eldo:
+        ifL(cond("isEqual", "person", input("meta.object")), [
+          ifL(cond("isEqual", "added", input("meta.action")),
+            set("createdByWebhook", true)
```
timliuhull: you don't need the "createdByWebhook" convention.  That was only for outreach.  it's superfluous in this case

packages/connectors/hull-pipedrive/server/glue.js
```diff
@@ -0,0 +1,264 @@
...
+      })
+    ])
+  ],
+  webhooks: ifL(input("body"), route("handleWebhook", cast(WebPayload, input("body")))),
+  handleWebhook:
+    ifL(cond("isEqual", "organization", input("meta.object")), {
+      do: hull("asAccount", input()),
+      eldo:
+        ifL(cond("isEqual", "person", input("meta.object")), [
+          ifL(cond("isEqual", "added", input("meta.action")),
```
timliuhull: if you want to do an "and" condition for an if, you can just put the condition in an array:
[cond(), cond()]

packages/connectors/hull-pipedrive/server/transforms-to-hull.js
```diff
@@ -0,0 +1,168 @@
...
+        outputPath: "${service_field_name}",
+        outputFormat: "${value.value}"
+      },
+      {
+        inputPath: "org_id",
+        outputPath: "hull_service_accountId",
+        outputFormat: "${value.value}"
+      },
+      {
+        condition: "createdByWebhook",
```
timliuhull: we can remove this too

PR #251, opened 2019-09-19
record logs and metrics

andyh565:


packages/hull/src/infra/instrumentation/metric-agent.js
```diff
@@ -28,13 +28,21 @@ class MetricAgent {
 
   metrics: Object;
 
+  recordedMetrics: any;
```
unity: Please avoid « any » and use a more exact type

packages/hull-client/src/client.js
```diff
@@ -467,11 +528,17 @@ class UserScopedHullClient extends EntityScopedHullClient {
    * @return {HullClient} HullClient scoped to a User and linked to an Account
    */
   account = (accountClaim: HullAccountClaims = Object.freeze({})) => {
-    return new AccountScopedHullClient({
-      ...this.config,
-      subjectType: "account",
-      accountClaim
-    });
+    const clientParams = {
+      recordedLogs: this.recordedLogs
```
unity: Shouldn’t this be included in the client config Instead of a separate parameter ?

packages/hull-client/src/client.js
```diff
@@ -91,9 +91,14 @@ class HullClient {
...
   batch: Firehose;
 
+  recordedLogs: Array<any>;
+
   static logger: HullClientStaticLogger;
 
-  constructor(config: HullClientConfig | HullClientCredentials) {
+  constructor(
+    config: HullClientConfig | HullClientCredentials,
+    clientParams?: Object
```
unity: Can’t the client Params be rolled up in the client config ?

timliuhull:
This was a prototype, we're proceeding differently, which is why i closed the pr

PR #264, opened 2019-09-26
fetch emails documentation

andyh565:


andyh565:
@julesb22 yeah it'll work if you hit Fetch Events button regardless of the toggle. 

Yeah, I can add a sentence about selecting the events

PR #275, opened 2019-10-03
Stop paging when timeOffset is older than lastFetchAt

louisjahn:


packages/connectors/hull-hubspot/server/lib/hubspot-client.js
```diff
@@ -312,14 +312,14 @@ class HubspotClient {
...
-    offset: ?string = null
+    offset: ?number = null
   ): Promise<HubspotGetAllContactsResponse> {
     return this.retryUnauthorized(() => {
       return this.agent
         .get("/contacts/v1/lists/recently_updated/contacts/recent")
         .query({
           count,
-          vidOffset: offset,
+          timeOffset: offset,
```
sbellity: @timliuhull @louisjahn looks like hubspot is recommending to use a combination of time and vid offset to page through the list whole list. Any reason for just using the timeOffset here ?

louisjahn: I tried every approach to make sure I wasn't missing anything, I did try using both `vidOffset` and `timeOffset`. We wanted to make sure that one of the edge cases would be suppressed with this update. Previously we only used `vidOffset` and we were experiencing this edge case.

That edge case is about a paging reset: When receiving a page, the `vidOffset` points towards the last user of the page. Hubspot will supposedly start paging starting from that user vid in the next page. Now imagine that this user receives an update before getting to the next page. You page through the users and receive the supposedly "next array of updated users". However, that user pointed out by `vidOffset` is the most recently updated, meaning that the next user in that page actually corresponds to the previous first user you received at the beginning of the request process. Result is that the paging got a reset.

During my tests, I switched to using both of these offsets. I still experienced that edge case. I then switched to only using `timeOffset` and the results were better: I kept scrolling through pages without getting a reset.

Yes this solution feels also weird to me. I may have been lucky during my tests so perhaps trying it out some more time wouldn't hurt.

timliuhull: We also just discussed the case (using vidoffset as well) where if Hull gets caught in a loop updating the same 100 contacts over and over, then the fetch recent contacts will also get caught in a loop as well with the same vids always going back to the top of the page

packages/connectors/hull-hubspot/server/purplefusion/hubspot-service.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+}
+
+const hubspotSdk: CustomApi = {
+  initialize: (context, api) => new HubspotSdk(context, api),
+  error: {
+    templates: [
+      {
+        truthy: { name: "ErrorTemplateMessage" },
```
timliuhull: What error is this for?  Does this error exist?

packages/connectors/hull-hubspot/server/purplefusion/service.js
```diff
@@ -155,6 +161,12 @@ const service: RawRestApi = {
         errorType: ConfigurationError,
         message: MESSAGES.STATUS_UNAUTHORIZED_ACCESS_TOKEN,
         recoveryroute: "refreshToken"
+      },
+      {
+        truthy: { status: 429 },
+        errorType: RateLimitError,
+        message: "Hubspot temporary API rate limit exceeded, retrying in a moment",
+        retry: 3
```
timliuhull: Need to change this to: "retryAttempts"

packages/connectors/hull-hubspot/server/purplefusion/hubspot-service.js
louisjahn: Talked with Andy about it, it's sort of the required template to make sure requests are all handled and not crash if one of them doesn't succeed.

packages/connectors/hull-hubspot/server/purplefusion/service.js
louisjahn: On it

PR #278, opened 2019-10-15
Entity deletions

andyh565:


packages/connectors/hull-hubspot/server/actions/status.js
```diff
@@ -71,10 +75,16 @@ async function statusCheckAction(ctx: HullContext): HullStatusResponse {
...
-  // Doesn't really matter to the customer I don't think
-  // if (!_.get(connector, "private_settings.portal_id")) {
-  //   pushMessage("Missing portal id.");
-  // }
+  const portalId = _.get(connector, "private_settings.portal_id", null);
+  const fetchDeleted =
+    _.get(connector, "private_settings.mark_deleted_contacts", false) ||
+    _.get(connector, "private_settings.mark_deleted_companies", false);
+  if (portalId !== null && fetchDeleted) {
+    const cachedCredentials = await getCachedClientCredentials(ctx, portalId);
```
timliuhull: I guess the idea here is that even though, we're getting portal id in the private settings, we're not actually using the credentials based off of a pull from the private settings? (which could be vulnerable to exploit where you could change your portal id to anything and get info from another portal).  This is only to check for a status message.  Would it be easier to just call: checkCachedCredentials which calls to the hubspot api to confirm the portal id anyway?

PR #300, opened 2019-11-06
Moving incoming success logs to debug level

sbellity:
Tests are failing, but I don't really understand why


timliuhull:
> Tests are failing, but I don't really understand why

Typeform tests are failing because they are expecting "debug" but receiving "info".  Did you forget to commit a change to the logging for typeform and only commit the change to the tests?

timliuhull:
@sbellity I fixed the merge conflicts with develop.  The remain test failures seem to be things where it's expecting "debug" but got "info".  Let me know if you've committed all of your code updates.  If you haven't I can go through and make the updates so that things go to debug instead of info, and it should fix the tests

PR #310, opened 2019-11-14
respond with flow control in account update

louisjahn:


timliuhull:
I think you could have set the FLOW_CONTROL_ACCOUNT_UPDATE_SUCCESS_SIZE too

FLOW_CONTROL_SIZE is deprecated

packages/connectors/hull-hubspot/server/handlers/account-update.js
```diff
@@ -9,5 +9,12 @@ export default async (
 ) => {
   const syncAgent = new SyncAgent(ctx);
   await syncAgent.sendAccountUpdateMessages(messages);
-  return {};
+  return {
+    flow_control: {
+      type: "next",
+      size: parseInt(process.env.FLOW_CONTROL_SIZE, 10) || 10,
+      in: parseInt(process.env.FLOW_CONTROL_IN, 10) || 5,
+      in_time: parseInt(process.env.FLOW_CONTROL_IN_TIME, 10) || 10
```
unity: As a best practice functions should be pure. You should pass those values as parameters from config -> handlers -> account_update. See https://github.com/hull/hull-connectors/blob/feature/processor/packages/connectors/hull-processor/server/config.js for an example

PR #313, opened 2019-11-15
Temporarily ignore configuration errors

louisjahn:
Temporarily ignore configuration errors while working on a permanent and better fix. Will reduce sentry calls.


timliuhull:
Any chance we can merge this to develop first?  Do we have cases that are throwing from develop that we can validate?

PR #328, opened 2019-11-25
Feature/outreach events

louisjahn:


packages/hull-connector-framework/src/purplefusion/transform-utils.js
```diff
@@ -10,39 +10,61 @@ function toUnixTimestamp() {
...
+      }
+    }
+    //evaluate transforma.validation.condition
+    //if true
+    // look at transform.validation.error
+    //if error === "breakeverything"
+  }
+}
+
+function toTransform(transform, context, input) {
```
timliuhull: The logic in here doesn't quite make sense to me.  Aren't the transform.condition, and the transform.validation completely different?  Couldn't it be more like:

```
evaluateValidation(transform, context, input);

  if (evaluateCondition(transform.condition, context, input)) {
    return true;
  }
```

PR #350, opened 2019-12-12
Feature/website connector

michaloo:
The website part connector which is important for now is just the manifest, documentation and logo. Everything else is in closed beta status and we will be testing it for CXL with Sven shortly.


unity:
@michaloo LGTM 🚀

PR #376, opened 2019-12-31
support meta fields on the contact

andyh565:


packages/connectors/hull-hubspot/server/lib/sync-agent/mapping-util.js
```diff
@@ -465,26 +465,35 @@ class MappingUtil {
...
+        let val = null;
+        if (_.startsWith(hsPropertyName, "contact_meta.")) {
+          const metaKey = hsPropertyName.split("contact_meta.")[1];
+          val = _.get(userData, metaKey, null);
+        } else if (properties && _.has(properties, hsPropertyName)) {
+          const propertiesKey = `properties[${hsPropertyName}].value`;
+          val = _.get(userData, propertiesKey, null);
+        }
+
+        if (!_.isNil(val)) {
```
timliuhull: I don't know the data well enough in this case, but will this keep us from getting values if they are "nulled" out?  because that's a valid case

andyh565: fixed

andyh565: it'll only operate on val if it isn't null, but still let null pass through

PR #395, opened 2020-01-14
support impersonation in vm connectors

unity:


unity:
@SMK1085 if you review this i can merge it. Allows (undocumented) access to `hull.asUser` and `hull.asAccount` from inside the Processor and Account Processor

PR #398, opened 2020-01-16
Log when setting null attributes

louisjahn:


michaloo:
@louisjahn can we possibly also log down the Hubspot service field name? Also, could we just pass those two names as the second argument to log function to make it maybe a little bit easier to build a report on?

PR #403, opened 2020-01-20
docker file; logging

andyh565:


Dockerfile
```diff
@@ -4,34 +4,55 @@ RUN apk add --no-cache rsync
...
+COPY packages/connectors/hull-marketo/package.json /app/packages/connectors/hull-marketo/package.json
+COPY packages/connectors/hull-outreach/package.json /app/packages/connectors/hull-outreach/package.json
+COPY packages/connectors/hull-pipedrive/package.json /app/packages/connectors/hull-pipedrive/package.json
+COPY packages/connectors/hull-processor/package.json /app/packages/connectors/hull-processor/package.json
+COPY packages/connectors/hull-scheduled-calls/package.json /app/packages/connectors/hull-scheduled-calls/package.json
+COPY packages/connectors/hull-slack/package.json /app/packages/connectors/hull-slack/package.json
 COPY packages/connectors/hull-typeform/package.json /app/packages/connectors/hull-typeform/package.json
+COPY packages/connectors/hull-warehouse/package.json /app/packages/connectors/hull-warehouse/package.json
+COPY packages/connectors/hull-website/package.json /app/packages/connectors/hull-website/package.json
+COPY packages/connectors/hull-zapier/package.json /app/packages/connectors/hull-zapier/package.json
```
sbellity: @andyh565 why are copying all those package.json files if we're only building the image for one $CONNECTOR ?

sbellity: Don't we want to build ONE docker image for the monorepo that includes all connector and leave it up to the container to start the one he wants ?

andyh565: we need to copy the package.json files over for the yarn workspaces

andyh565: I'm fine if we want to build all the connectors in one image

PR #435, opened 2020-02-06
Feature/coppercrm

timliuhull:
Pretty safe for the most part, contains many new features and the copper connector itself.  I've noted the 2 places I need to double check


packages/hull-connector-framework/src/purplefusion/variable-context.js
```diff
@@ -30,13 +31,59 @@ class HullVariableContext {
...
+      const currentContext = this.localContext[i];
+      const value = _.get(currentContext, key);
       if (value !== undefined) {
         return value;
       }
+
+      const keys = Object.keys(currentContext);
+      if (keys.indexOf(rootKey) >= 0) {
+        // console.log(`Returning undefined because root key is present in context: ${rootKey} for ${key}`);
+        return value;
```
timliuhull: Returning the value if the variable was found, even if it was null or undefined.  we don't want to fall back to previous declarations if the variable was called here

timliuhull: Committed this pr to log this scenario before we go to prod: https://github.com/hull/hull-connectors/pull/434

packages/hull/src/middlewares/credentials-from-query.js
```diff
@@ -79,6 +79,8 @@ function credentialsFromQueryMiddlewareFactory() {
         parseToken(req.hull.clientCredentialsEncryptedToken, hostSecret) ||
         parseToken(req.hull.clientCredentialsToken, hostSecret) ||
         parseToken(getToken(req.query), hostSecret) ||
+        // also check the body to see if the tokens may be there
+        parseToken(getToken(req.body), hostSecret) ||
```
timliuhull: We should confirm that this is an ok and safe addition to the "credentials-from-query" pattern

PR #437, opened 2020-02-07
Hotfix/clearbit roles

unity:


packages/connectors/hull-clearbit/manifest.json
```diff
@@ -1609,7 +1609,7 @@
       "description": "Hit `Enter` after every title to make a Tag out of it, Restrict your search by specifying criteria for the prospect's title. If left blank, search is carried out without restriction."
     },
     {
-      "name": "prospect_filter_role",
+      "name": "prospect_filter_roles",
       "title": "Filter for a certain Role",
```
SMK1085: Can we make the title plural as well please

packages/connectors/hull-clearbit/manifest.json
```diff
@@ -1644,7 +1644,7 @@
       "description": "Specify a certain role or leave blank to search without restriction."
     },
     {
-      "name": "prospect_filter_seniority",
+      "name": "prospect_filter_seniorities",
       "title": "Filter for specific Seniority",
```
SMK1085: Same here, can we have plural please

unity: Fixed

packages/connectors/hull-clearbit/manifest.json
unity: Fixed

PR #438, opened 2020-02-07
Backmerge release

louisjahn:


louisjahn:
Released in this PR:

- Hull-repl connector
- Send users and accounts segments as attributes to outreach
- CircleCI build, publish and deploy docker images of connectors

PR #441, opened 2020-02-10
refactored a little added unit tests and fixed a couple issues

timliuhull:


packages/connectors/hull-outreach/server/transforms-to-service.js
```diff
@@ -99,23 +128,15 @@ const transformsToService: ServiceTransforms = [
...
-          {
-            arrayStrategy: "json_stringify",
-            condition: notNull("connector.private_settings.prospect_outgoing_account_segments"),
-            inputPath: "flattened_account_segments",
-            outputPath: "data.attributes.${connector.private_settings.prospect_outgoing_account_segments}",
-          }
+          createStringifyTransform("flattened_segments", "connector.private_settings.prospect_outgoing_user_segments", outreachProspectArrayFields),
+          createStringifyTransform("flattened_account_segments", "connector.private_settings.prospect_outgoing_account_segments", outreachProspectArrayFields),
+          createRawArrayTransform("flattened_segments", "connector.private_settings.prospect_outgoing_user_segments", outreachProspectArrayFields),
+          createRawArrayTransform("flattened_account_segments", "connector.private_settings.prospect_outgoing_account_segments", outreachProspectArrayFields)
```
timliuhull: Primary issue was having to differentiate where we wanted to stringify (string fields) or send as raw array (array fields).  We have a better way of doing this in the new framework with schemas, but this should fix this up for now.  Added tests where we can refactor and have test coverage in the future

packages/connectors/hull-outreach/server/transforms-to-service.js
```diff
@@ -21,8 +50,8 @@ const transformsToService: ServiceTransforms = [
             expression: "$merge([\n" +
               "    $spread(),\n" +
               "    {\n" +
-              "        \"flattened_segments\" : segments.name,\n" +
-              "        \"flattened_account_segments\": account_segments.name\n" +
+              "        \"flattened_segments\" : [segments.name],\n" +
+              "        \"flattened_account_segments\": [account_segments.name]\n" +
```
timliuhull: Needed to wrap in braces because in jsonata, if not wrapped in brances, a 1 value array will resolve to a string rather than a 1 value array

packages/connectors/hull-outreach/server/transforms-to-service.js
```diff
@@ -134,7 +155,7 @@ const transformsToService: ServiceTransforms = [
             expression: "$merge([\n" +
               "    $spread(),\n" +
               "    {\n" +
-              "        \"flattened_account_segments\" : account_segments.name\n" +
+              "        \"flattened_account_segments\" : [account_segments.name]\n" +
```
timliuhull: Needed to wrap in braces because in jsonata, if not wrapped in brances, a 1 value array will resolve to a string rather than a 1 value array

packages/connectors/hull-outreach/server/transforms-to-service.js
unity: You could more elegantly do `segments.name[]` - would result in an array 100% of the time

unity: http://docs.jsonata.org/next/predicate#singleton-array-and-value-equivalence - If this is a concern, then the expression can be modified to make it return an array even if only a single value is matched. This is done by adding empty square brackets [] to a step within the location path. The examples above can be re-written to always return an array as follows:

packages/connectors/hull-outreach/server/transforms-to-service.js
unity: @timliuhull i suggest using the syntax above which is the native way to cast as array.
I've used this in the attribute mapper for the same result:

```
[
  { "hull": "segments.name", "service": "hull_segments", "castAs": "array" }
]
```

https://github.com/hull/hull-connectors/blob/13471133447c087cb03a2e6687777e6855126fc1/packages/hull/src/helpers/map-attributes.js#L12

timliuhull: already done

PR #451, opened 2020-02-13
Release/coppercrm merge

timliuhull:
reviewed Copper, Warehouse and Outreach updates.


packages/connectors/hull-pipedrive/package.json
```diff
@@ -1,10 +1,9 @@
 {
   "name": "hull-pipedrive",
-  "version": "0.1.17-monorepo",
+  "version": "0.1.18-monorepo",
   "description": "Sync persons and organizations with Hull",
   "scripts": {
-    "test": "yarn run test:integration",
-    "test:integration": "DEBUG=hull* NODE_ENV=test mocha --exit --require ../../../root-babel-register.js -R spec ./test/integration/{*.js,**/*.js}",
+    "test": "exit 0; echo 'tests are run by global jest'",
```
timliuhull: I think this is right... should be running pipedrive tests with jest

packages/hull-client/src/client.js
```diff
@@ -199,13 +234,45 @@ class HullClient {
...
-        logsArray.push({
-          message,
-          level,
-          data: payload.data,
-          context: payload.context,
-          timestamp: new Date().toISOString()
-        });
+        let transformedLog;
+
+        if (this.config.esLogTransform) {
```
timliuhull: @andyh565 This look right?

packages/connectors/hull-pipedrive/package.json
louisjahn: Yes they are!

packages/hull-client/src/client.js
andyh565: yeah if you're good with it

PR #452, opened 2020-02-18
Feature/ssh datawarehouse

andyh565:


packages/connectors/hull-warehouse/server/postgres-sequalize-service.js
```diff
@@ -411,36 +485,39 @@ class SequalizeSdk {
...
+                  id: previous
+                }
+              });
+          })
+        });
+    });
   }
 
   async removeHullAccount(id: String) {
-    return this.getSequelizeConnection()
```
timliuhull: ouch... was this not working before?

packages/connectors/hull-warehouse/server/postgres-sequalize-service.js
```diff
@@ -411,36 +485,39 @@ class SequalizeSdk {
       merged: String
     }
   ) {
-    return this.getSequelizeConnection()
```
timliuhull: did we change this because it wasn't working?  I feel like this worked.

PR #462, opened 2020-02-24
website on kafka

sbellity:


packages/connectors/hull-website/server/config.js
```diff
@@ -7,31 +7,58 @@ export default function connectorConfig(): HullConnectorConfig {
...
     hostSecret,
-    devMode: NODE_ENV === "development",
+    devMode: DISABLE_DEV_MODE ? false : NODE_ENV === "development",
     port: PORT || 8082,
     handlers: handlers({
-      redisUri: REDIS_URL
+      redisUri: REDIS_URL,
+      HULL_DOMAIN,
+      REMOTE_DOMAIN,
+      firehoseTransport: {
```
unity: Shouldn’t this be generic and passed through clientConfig instead of passed to the handler ?

packages/connectors/hull-website/server/handlers/traits-handler.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+export default (req, res) => {
+  const { name, operation, value } = req.body;
+  return req.hull
+    .traits({ [name]: { operation, value } })
+    .then(
```
unity: You might want to use a async/await as a good practice

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,195 @@
...
+      };
+      if (config.userClaim) claims["io.hull.asUser"] = config.userClaim;
+      if (config.accountClaim)
+        claims["io.hull.asAccount"] = config.accountClaim;
+      message.claims = claims;
+    }
+
+    const producerName = config.connectorName || "";
+
+    return getProducer(firehoseTransport, producerName).then(
```
unity: Async/await would be clearer. Especially to avoid un caught promises.

packages/connectors/hull-website/server/config.js
```diff
@@ -7,31 +7,58 @@ export default function connectorConfig(): HullConnectorConfig {
...
+  if (!SECRET && NODE_ENV !== "development") {
+    throw new Error("Missing SECRET environment variable");
+  }
+
   const hostSecret = SECRET || "1234";
   return {
     manifest,
     hostSecret,
-    devMode: NODE_ENV === "development",
+    devMode: DISABLE_DEV_MODE ? false : NODE_ENV === "development",
```
unity: @sbellity - this is super weird. Can we try to avoid this before merging ?

packages/connectors/hull-website/server/config.js
sbellity: Yes for all connectors except for this one because it creates it own instance of hull client in the handler.

packages/connectors/hull-website/server/config.js
sbellity: done

PR #466, opened 2020-02-27
client with session config

sbellity:


packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+        const timeoutError = new Error();
+        timeoutError.message = "Backend connection timeout";
+        return Promise.reject(timeoutError);
+      }
+    );
+  };
+}
+
+function flushProducer(brokers, { producer, ready }) {
+  return ready.then(() => {
```
unity: Not catching errors here due to dangling promises. Suggest to use try/catch+async-await pattern to avoid silent errors.

packages/hull-client/src/client.js
```diff
@@ -95,7 +91,13 @@ class HullClient {
 
   static logger: HullClientStaticLogger;
 
-  constructor(config: HullClientConfig | HullClientCredentials) {
+  static exit() {
+    return Promise.all([Firehose.exit(), FirehoseKafka.exit()]).then(flushed =>
```
unity: Not catching errors here due to dangling promises. Should use ˋasync static exit` and 

```
try {
  const flushed = await Promise.all(...)
} catch (err) {
  //do something / log something
}
```

PR #467, opened 2020-02-27
firehose with context

sbellity:


packages/connectors/hull-website/server/handlers/index.js
```diff
@@ -12,13 +16,21 @@ import onConnectionFactory from "../lib/on-connection";
...
-): HullHandlersConfiguration => {
-  const { server, Client, getContext } = connector;
+const handlers = ({
+  redisUri,
+  firehoseTransport,
+  HULL_DOMAIN,
+  REMOTE_DOMAIN
+}: {
+  redisUri: string,
+  firehoseTransport: HullFirehoseKafkaTransport
```
unity: Can you add Flow types to `REMOTE_DOMAIN` and `HULL_DOMAIN`? From what I see we don't handle Undefined values very clearly

packages/connectors/hull-website/server/handlers/remote-handler.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    res.cookie("_sid", sessionId, {
+      secure: true,
+      sameSite: "None",
+      maxAge: THIRTY_MINUTES,
+      httpOnly: true
+    });
+
+    fetchRemoteConfig(CACHE, req.organization, appId).then(
```
unity: Errors in `renderRemote` wouldn't be catched here - can we use async-await with a try/catch instead?

packages/connectors/hull-website/server/handlers/track-handler.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+export default (req, res) => {
+  const { event, properties } = req.body;
+  return req.hull
```
unity: can we use async-await with a try-catch for code consistency here?

packages/connectors/hull-website/server/handlers/traits-handler.js
```diff
@@ -0,0 +1,28 @@
...
+        ? { [name]: value }
+        : { [name]: { operation, value } };
+  }
+
+  if (traits.email && EMAIL_REGEXP.test(traits.email)) {
+    const anonymous_id = req.get("hull-bid") || req.cookies._bid;
+    req.hull = req.hull.asUser({ email: traits.email, anonymous_id });
+  }
+
+  return req.hull
```
unity: can we use async-await with a try-catch for code consistency here?

packages/connectors/hull-website/server/handlers/track-handler.js
```diff
@@ -0,0 +1,11 @@
...
+    .track(event, properties, req.firehoseEventContext)
+    .then(
+      ok => res.status(204).send({ ok: !!ok }),
```
unity: I would abstract the responses in an external helper given the fact they're identical across different handlers. Just so that, should we need to fix them we do it all in the same place.

packages/hull-client/src/client.js
```diff
@@ -95,7 +92,13 @@ class HullClient {
 
   static logger: HullClientStaticLogger;
 
-  constructor(config: HullClientConfig | HullClientCredentials) {
+  static exit() {
+    return Promise.all([Firehose.exit(), FirehoseKafka.exit()]).then(flushed =>
```
unity: no catching of errors here, dangling promise - need to use async/await with a try-catch instead

packages/hull-client/src/client.js
```diff
@@ -137,20 +140,34 @@ class HullClient {
...
+            timeout,
+            retry,
+            domain = "",
+            protocol = "",
+            firehoseUrl = `${protocol}://firehose.${domain}`
+          } = clientConfig;
+          return restAPI(this, batcher.config, firehoseUrl, "post", params, {
+            timeout,
+            retry,
+            isFirehose: true
```
unity: Can you document this new param somewhere?

packages/hull-client/src/lib/configuration.js
```diff
@@ -41,17 +42,23 @@ const VALID = {
...
 const REQUIRED_PROPS = {
   id: VALID.objectId,
-  secret: VALID.string,
   organization: VALID.string
 };
 
 const VALID_PROPS = {
   ...REQUIRED_PROPS,
+  secret: VALID.string,
+  trackingOnly: VALID.boolean,
```
unity: Can you document the format of this param in `NEW_README` ?

packages/hull-client/src/lib/configuration.js
```diff
@@ -69,16 +76,17 @@ const VALID_PROPS = {
   connectorName: VALID.string,
   requestId: VALID.string,
   logs: VALID.array,
-  firehoseEvents: VALID.array
+  firehoseEvents: VALID.array,
```
unity: Can you document the format of this param in `NEW_README` ?

packages/hull-client/src/lib/configuration.js
```diff
@@ -69,16 +76,17 @@ const VALID_PROPS = {
...
+  firehoseTransport: VALID.transport
```
unity: Can you document the format of this param in `NEW_README` ?

packages/hull-client/src/lib/configuration.js
```diff
@@ -41,17 +42,23 @@ const VALID = {
   },
   array(arr) {
     return _.isArray(arr);
+  },
+  transport(t) {
```
unity: shouldn't this method be name `kafkaTransport` instead? it seems it's only returning true if the transport is a valid kafkaTransport

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+
+import type {
+  HullClientInstanceConfig,
+  HullClientLogger,
+  HullFirehoseKafkaTransport,
+  HullUserClaims,
+  HullAccountClaims
+} from "../types";
+
+type FirehoseMessage = {
```
unity: Shouldn't the Types in this file be generic and not specific to Kafka Transport ? If so can we move them to the `types` and name them `HullFirehoseMessage` etc ... ?

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+}
+
+function getProducer(
+  transport: HullFirehoseKafkaTransport,
+  connectorName: string
+) {
+  const brokers = transport.brokersList.join(",");
+  PRODUCERS[brokers] =
+    PRODUCERS[brokers] || buildProducer(transport, connectorName);
+  return PRODUCERS[brokers].ready;
```
unity: This whole section with buildProducer is not very clear.
Can we try to de-nest it with something like

```
async function buildProducer(){
  ...
  return new Promise((resolve,reject)=>{...});
}
function getProducer(){
  PRODUCERS[brokers] = PRODUCERS[brokers] || buildProduce(...)
}
async function getInstance(){
  try {
    const producer = await getProducer(firehoseTransport, producerName)
    ....
  } catch (err){
  ...
  }
}

```

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+      if (config.accountClaim)
+        claims["io.hull.asAccount"] = config.accountClaim;
+      message.claims = claims;
+    }
+
+    const producerName = config.connectorName || "";
+
+    return getProducer(firehoseTransport, producerName).then(
+      producer => {
+        return new Promise((resolve, reject) => {
```
unity: the error flow isn't clearly readable due to the nesting of promises. I suggest to flatten this.

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+                clientError.message = "Backend publication error";
+                return reject(clientError);
+              }
+              return resolve({ ok: true, offset });
+            }
+          );
+        });
+      },
+      () => {
+        const timeoutError = new Error();
```
unity: Rejected errors from the `producer.produce` call won't be catched here.

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+        timeoutError.message = "Backend connection timeout";
+        return Promise.reject(timeoutError);
+      }
+    );
+  };
+}
+
+function flushProducer(brokers, { producer, ready }) {
+  return ready.then(() => {
```
unity: can we await instead of nesting promises here?

```
async function flushProducer(brokers, producerInstance){
    const producer = await producerInstance();
    return new Promise((resolve, reject) => {
      producer.flush(30000, err => {
        if (err){ reject(err) }
        resolve(`xxx`)
      });
    });
  }
}
```

packages/hull-client/src/lib/firehose-kafka.js
```diff
@@ -0,0 +1,196 @@
...
+  });
+}
+
+function exit() {
+  IS_EXITING = true;
+  const brokersLists = Object.keys(PRODUCERS);
+  const flushes = brokersLists.map(brokers =>
+    flushProducer(brokers, PRODUCERS[brokers])
+  );
+  return Promise.all(flushes);
```
unity: This isn't explicitely handling errors - can we await this and `try-catch` it ?

packages/hull-client/src/lib/firehose.js
```diff
@@ -6,6 +6,18 @@ const BATCHERS = {};
...
+  static exit() {
+    FirehoseBatcher.exiting = true;
+    const cb = k => (err, ok) => {
+      const msg = `Flushed batcher ${k} -- ok=${ok} -- err=${err}`;
+      return msg;
+    };
+    const flushed = _.map(BATCHERS, (b, k) => {
+      b.flush(cb(k));
+    });
+    return Promise.all(flushed);
```
unity: This isn't explicitely handling errors

packages/hull-client/src/lib/firehose.js
```diff
@@ -51,7 +63,7 @@ class FirehoseBatcher {
 
   flush(fn) {
     fn = fn || (() => {});
-    if (!this.queue.length) return setImmediate(fn);
+    if (!this.queue.length) return Promise.resolve(fn);
```
unity: if you make `flush` an `async` you don't need to do this. You can just return the `fn`

packages/hull-client/src/lib/firehose.js
```diff
@@ -63,18 +75,13 @@ class FirehoseBatcher {
       sentAt: new Date()
     };
 
-    return this.handler(params, this).then(
+    const flushed = this.handler(params, this);
+    flushed.then(
       ok => fns.forEach(func => func(null, ok)),
       err => fns.forEach(func => func(err, null))
     );
+    return flushed.then(fn, fn);
```
unity: This swallows errors in each `fns`

packages/hull/src/connector/hull-connector.js
```diff
@@ -229,8 +235,16 @@ class HullConnector {
     }
     if (this.serverConfig.start) {
       const app = express();
+
+      if (this.connectorConfig.trustProxy) {
```
unity: can you document thoroughly `trustProxy` and what it does in `NEW_README` as it's going to be evaluated in all connectors ?

Dockerfile
```diff
@@ -15,39 +25,37 @@ COPY babel.config.js /app/
...
-CMD rsync -mrv --include="*/" --include="package.json" --exclude="*" ./ /app/
+COPY packages/hull/package.json /app/packages/hull/
+COPY packages/hull-client/package.json /app/packages/hull-client/
+COPY packages/hull-connector-framework/package.json /app/packages/hull-connector-framework/
+COPY packages/hull-vm/package.json /app/packages/hull-vm/
+COPY packages/hullrepl/package.json /app/packages/hullrepl/
 
-COPY ./ /app/
+# CONNECTORS:
+COPY packages/connectors/hull-account-processor/package.json /app/packages/connectors/hull-account-processor/package.json
```
unity: Seems like a huge hassle and reason for incomprehension and obscure issues waiting to happen here - can't we avoid manually copying the files ?

package.json
```diff
@@ -50,24 +52,8 @@
     "yarn": "^1.10.x"
   },
   "devDependencies": {
-    "@babel/cli": "^7.4.4",
```
unity: I think I remember Heroku needs the dependencies in `dependencies` - meaning this build might not deploy on Heroku

packages/connectors/hull-website/server/handlers/legacy-v1-api-compatibility.js
```diff
@@ -0,0 +1,120 @@
...
+          "hull-sid"
+        ];
+        if (
+          originHost === `${req.orgNamespace}.${HULL_DOMAIN}` ||
+          originHost === `${req.orgNamespace}.${REMOTE_DOMAIN}`
+        ) {
+          return callback(null, { origin: true, allowedHeaders });
+        }
+
+        return callback(new UnauthorizedDomainError(req.organization), {
```
unity: Are you sending the same detailed error messages as in the legacy system ? We want customers to have explicit explanations if their library doesn't load.

Dockerfile
sbellity: Apparently no. A lot of people are also complaining about this but I can confirm that the rsync thing does not work in this context.

packages/connectors/hull-website/server/handlers/legacy-v1-api-compatibility.js
sbellity: For this one, it will be a CORS error surfaced by the browser

package.json
sbellity: I have checked and it build correctly on Heroku with that package.json

timliuhull:
@unity Are all of those changes necessary?  Maybe we should let the connector team give feedback on how they'd like to test this and the patterns they'd prefer to implement since they are the owners of this code

packages/connectors/hull-website/server/handlers/legacy-v1-api-compatibility.js
unity: Which isn't what the previous setup did. It was customer friendly and exposed the valid domains. We did that after realizing customers didn't understand the system error messages. Please make sure we have feature parity

unity:
> @unity Are all of those changes necessary? Maybe we should let the connector team give feedback on how they'd like to test this and the patterns they'd prefer to implement since they are the owners of this code

I think they are, if we want consistency in the code conventions with the existing implementation. At the framework level, there is complexity in the code and dataflows that can generate uncaught errors

packages/connectors/hull-website/server/handlers/legacy-v1-api-compatibility.js
sbellity: Domain validation is now (and has been since str8) via the relatively safer CSP header: https://github.com/hull/hull-connectors/blob/feature/firehose-with-context/packages/connectors/hull-website/server/handlers/remote-handler.js#L41

The hidden iframe in the case of non whitelisted domain is not even evaluated by the browser, which displays the following error message in the console: 

Local_dev

packages/connectors/hull-website/server/handlers/index.js
sbellity: at runtime config.js throws an error if they are not set. I think we're good

PR #482, opened 2020-03-05
Outgoing webhook fixes, and standardize hasMatchingTriggers and hasMa…

unity:
…tchingSegment helpers


packages/hull-connector-framework/src/purplefusion/triggers/validations.js
```diff
@@ -1,91 +1,119 @@
...
+): boolean %checks => !_.isEmpty(_.intersection(whitelist, _.keys(changes)));
+
+// TODO: generate global whitelist from whitelist
+const validateEvents: HullTriggerValidationFunction = (
+  events: Array<HullEvent>,
+  whitelist: Array<string>
+): boolean %checks =>
+  !_.isEmpty(_.intersection(whitelist, _.map(events, "event")));
+
+// TODOANDY -> is this a custom entry? We already use `ALL` as a value everywhere else.
```
andyh565: @unity I'll update Zapier and migrate the settings to use "ALL" instead of "all_segments"

PR #492, opened 2020-03-12
Support uuid and phone library in Processor

unity:
@michaloo We need to check for Performance impact of freezing such a large library (google-libphonenumber) - if impact is confirmed we will need to pull back and think of better ways.


michaloo:
@unity is it freezing the libraries when the connector process starts or whenever a notification is processed?

unity:
> @unity is it freezing the libraries when the connector process starts or whenever a notification is processed?

It's freezing on each notification.

Initial, quick Perf testing showed incredibly fast performance for VM compared to the entire request cycle, We would need real world profiling to see if it's actually a bottleneck (not sure it is)

PR #500, opened 2020-03-18
ssh documentation

andyh565:


michaloo:
@andyh565 is there any specific about any of those fields? I understand that I can leave port empty to default to 22. What about the private key? I assume I just paste the key string, but should I be aware of any formatting requirements?

andyh565:
> @andyh565 is there any specific about any of those fields? I understand that I can leave port empty to default to 22. What about the private key? I assume I just paste the key string, but should I be aware of any formatting requirements?

There's nothing specific or hidden regarding these fields. The private key can be pasted in as it is in the private key file

PR #503, opened 2020-03-19
Implement New Jsonata Attribute Mapper and Expose Hubspot Default Mapping

andyh565:


packages/connectors/hull-hubspot/server/config.js
```diff
@@ -5,6 +5,8 @@ import manifest from "../manifest.json";
 import handlers from "./handlers";
 import hubspotWebhookHandler from "./handlers/hubspot-webhook-handler";
 
+require("dotenv").config();
```
andyh565: TODO remove

PR #508, opened 2020-03-30
Release/processor UUID phone

louisjahn:


unity:
LGTM - One thing to monitor very closely is the performance of this release for the Processor and Account processor deployments. We're serializing larger objects, we want to monitor behavior

louisjahn:
> LGTM - One thing to monitor very closely is the performance of this release for the Processor and Account processor deployments. We're serializing larger objects, we want to monitor behavior

I'll be monitoring both metrics

andyh565:
@louisjahn LGTM except, run the script 

`./scripts/version-bump.sh 0.1.23-monorepo`

to update all manifests and package.json versions

PR #519, opened 2020-04-13
Release/0.1.24 monorepo

andyh565:


packages/connectors/hull-sql/assets/index.html
```diff
@@ -0,0 +1,11 @@
+<html>
+<head>
+  <title>Segment Ship</title>
```
timliuhull, line 3: Was this in the sql connector too?

webpack/webpack.config.js
```diff
@@ -58,6 +58,10 @@ const buildConfig = ({ assets, files, destination, mode = "production" }) => ({
 
   module: {
     rules: [
+      {
+        test: /\.css$/,
+        use: [{ loader: "style-loader" }, { loader: "css-loader" }]
+      },
```
timliuhull, line 64: a bit out of the way for this change, was this intended?

packages/connectors/hull-sql/assets/index.html
andyh565: ah yeah I didn't catch that. It was in the original sql connector

webpack/webpack.config.js
andyh565: I'll double check this. It may have been left over from when I was trying to get webpack to work

unity:
Since we're gonna have to QA this - would it be valuable to move the import strategy to the Kakfa adapter at the same time ?

andyh565:
> Since we're gonna have to QA this - would it be valuable to move the import strategy to the Kakfa adapter at the same time ?

That makes sense to me. @timliuhull what do you think?

unity:
It would make it possible to support the session ID capture and other things we don't support via the import API without needing platform-side changes.

timliuhull:
> > Since we're gonna have to QA this - would it be valuable to move the import strategy to the Kakfa adapter at the same time ?
> 
> That makes sense to me. @timliuhull what do you think?

I think the Import Api needs to be able to handle session_id etc... Unless we think the Import Api is not the way we want to do bulk imports in the future.  

Kafka adapter was a replacement for the firehose, not imports.  I don't know the details, but I'd think a bulk import endpoint would always be useful, especially in cases where we're pulling stuff directly from a database like this and the volume could be pretty large.

And just as a general instinct, I'd try not to put too many things together and make an already big story bigger.

andyh565:
> > > Since we're gonna have to QA this - would it be valuable to move the import strategy to the Kakfa adapter at the same time ?
> > 
> > 
> > That makes sense to me. @timliuhull what do you think?
> 
> I think the Import Api needs to be able to handle session_id etc... Unless we think the Import Api is not the way we want to do bulk imports in the future.
> 
> Kafka adapter was a replacement for the firehose, not imports. I don't know the details, but I'd think a bulk import endpoint would always be useful, especially in cases where we're pulling stuff directly from a database like this and the volume could be pretty large.
> 
> And just as a general instinct, I'd try not to put too many things together and make an already big story bigger.

Yeah I think this original story is closed and we should be able to merge it and anything else will need to be additional stories. There's no risk in merging this as is as no one will be using it right now anyway

webpack/webpack.config.js
andyh565: @timliuhull yes it's needed for the package codemirror

PR #523, opened 2020-04-15
pass in check keys option to allow "." in keys

andyh565:
…_Model.create

### DIFF
--- a/packages/hull-vm/server/save-recent.js
+++ b/packages/hull-vm/server/save-recent.js
@@ -26,5 +26,7 @@ export default function saveRecent(
     payload,
     date: new Date().toString()
   };
-  return EntryModel.create(entry);
+  return EntryModel.create([entry], {
+    checkKeys: false
+  });
 }


andyh565:
documentation: https://mongoosejs.com/docs/api.html#model_Model.create

packages/hull-vm/server/save-recent.js
```diff
@@ -26,5 +26,7 @@ export default function saveRecent(
     payload,
     date: new Date().toString()
   };
-  return EntryModel.create(entry);
+  return EntryModel.create([entry], {
```
andyh565, line 29: entry must be an array according to the documentation:
https://mongoosejs.com/docs/api.html#model_Model.create

unity:
Seems ok to me. Haven’t tested the behavior or side effects doing this. But it seems like a good straightforward fix

packages/hull-vm/server/save-recent.js
```diff
@@ -26,5 +26,7 @@ export default function saveRecent(
...
+    checkKeys: false
+  });
```
andyh565, line 31: entry must be an array according to the documentation:
https://mongoosejs.com/docs/api.html#model_Model.create

unity: It seemed to work before - do we know why ?

andyh565:
> Seems ok to me. Haven’t tested the behavior or side effects doing this. But it seems like a good straightforward fix

okay great. The fix may be a little more complicated for the legacy connector since I'll need to update the mongoose version as well

packages/hull-vm/server/save-recent.js
andyh565: I'm not sure. I was able to reproduce the error the first time by adding a "." in one of the keys

andyh565: > It seemed to work before - do we know why ?

yes, it only needs to be an array if you pass in the options

PR #531, opened 2020-04-23
optional cache prefix

andyh565:


packages/hull/src/index.js
```diff
@@ -134,7 +135,8 @@ const buildConfigurationFromEnvironment = env => {
   const cacheConfig = {
     ...cacheAdapter,
     ttl: SHIP_CACHE_TTL || 60,
-    max: SHIP_CACHE_MAX || 100
+    max: SHIP_CACHE_MAX || 100,
+    cachePrefix: SHIP_CACHE_PREFIX
```
sbellity, line 139: I would simply call this key `prefix` since we're already in the context of the cacheConfig. Or... maybe `keyPrefix` if you want something more explicit

packages/hull/src/infra/cache/connector-cache.js
```diff
@@ -44,6 +52,9 @@ class ConnectorCache {
         "ConnectorCache can be used only with initialized client, otherwise use ctx.cache.cache.set"
       );
     }
+    if (this.cachePrefix) {
+      return `${this.cachePrefix}${key}`;
```
sbellity, line 56: I would also add a separator to make it easier to read the keys in a redis client... maybe something like : 

```return `${this.cachePrefix}:${key}`;```

packages/hull/src/index.js
andyh565: done

PR #532, opened 2020-04-23
use connector id in cache key

andyh565:


packages/hull/src/infra/cache/connector-cache.js
```diff
@@ -52,10 +52,10 @@ class ConnectorCache {
         "ConnectorCache can be used only with initialized client, otherwise use ctx.cache.cache.set"
       );
     }
+    const { id, secret, organization } = this.ctx.client.configuration();
     if (this.keyPrefix) {
-      return `${this.keyPrefix}:${key}`;
+      return `${this.keyPrefix}:${id}:${key}`;
```
timliuhull, line 57: One idea that might not be horrible is to try and use the same key strategy as the jwt encode below... using secret?  that way we could at least ensure it's functionally pretty similar to the previous implementation, wouldn't necessarily rely on the "keyPrefix" or something...

timliuhull: But I guess you don't want the secret in plain text in the cache key...

timliuhull: Wouldn't the id be sufficient? I guess I would then ask the question, what is the purpose of the keyPrefix.

timliuhull: @sbellity

unity: I think "key" already contains the Ship ID

https://github.com/hull/hull-connectors/blob/05022aab76a98f53c93f1d81694d4c18c2711a16/packages/hull/src/infra/cache/connector-cache.js#L49

sbellity: The purpose of this key prefix is allowing different connectors to share a redis cluster. It's there to act as a connector specific namespace on the global Redis keyspace.

PR #536, opened 2020-04-30
Feature/coppercrm outgoing leads

timliuhull:


packages/hull-connector-framework/src/purplefusion/router.js
```diff
@@ -134,27 +134,58 @@ class HullRouter {
...
             dataToSkip.push(message);
           }
         });
+
+        // Right now this logic isn't quite right.  For batch scenarios, we'll default to sending leads if the user
+        // isn't in any segment.  This is ok for now, but ask product in the future if we want to change
+        // salesforce connector doesn't send at all if user is in neither segment, which might be ok
+        // we would have to extract the "send regardless" logic out to here, but that's probably where it belongs anyway
+        if (entityType === "user") {
+          if (!isUndefinedOrNull(_.get(context, "connector.private_settings.synchronized_lead_segments"))) {
```
timliuhull, line 159: This is the logic for sending leads if they are in a synchronized lead segment.  Implemented as salesforce does it currently.  This logic will not be triggered at all unless you have a synchronized_lead_segments attribute in your manifest, so it should be pretty sage

packages/hull-connector-framework/src/purplefusion/router.js
```diff
@@ -134,27 +134,58 @@ class HullRouter {
...
         dataToSend = data;
       }
 
+
       // I like getting rid of the splitting here, and always passing in the data
       // glue can handle the split if needed
       // TODO need to test sending an empty array, or decide if we need extra logic here...
       let dispatchPromise = dispatcher.dispatchWithData(context, route, objectType, dataToSend);
 
+      if (!_.isEmpty(leadDataToSend)) {
```
timliuhull, line 183: only do something if leadDateToSend is not empty, which should only be if you have a lead_synchronized_segments, which no manifests running purple fusion currently do

packages/hull-connector-framework/src/purplefusion/utils.js
```diff
@@ -20,13 +20,19 @@ async function asyncForEach(toIterateOn, asyncCallback) {
...
   }
 
-  for (let i = 0; i < toIterateOn.length; i += 1) {
-    await asyncCallback(toIterateOn[i], i);
+  if (Array.isArray(toIterateOn)) {
+    for (let i = 0; i < toIterateOn.length; i += 1) {
+      await asyncCallback(toIterateOn[i], i);
+    }
+  } else {
+    const keys = Object.keys(toIterateOn);
```
timliuhull, line 32: Committed a log if there is currently any scenarios where "toIterateOn" is an object where we could be changing the behavior, and nothing came up. So this should be safe

https://github.com/hull/hull-connectors/pull/539

PR #537, opened 2020-04-30
Remaining api calls for Pipedrive

louisjahn:


michaloo:
This solution didn't work, we have created a backlog story to implement it correctly (requires contacting Pipedrive support).

PR #546, opened 2020-05-07
Update docs for new webhooks setup procedure

louisjahn:


michaloo:
We discussed having both - old and new - instructions at the same time in order to support customers on old and new c.io instances (created before and after April 8th 2020)

PR #566, opened 2020-05-13
default to empty list for attributes in private settings

andyh565:


andyh565:
> Sure, do you want to do it with the other attributes in the list too? "incoming_user_claims", "incoming_account_claims"? Maybe that wouldn't happen because they have default values, and even if removed, it should have an array?

yeah you're probably right. Right now, all of the attributes will have a default value, whereas before it didn't. The claims always had a default value

PR #577, opened 2020-05-16
cache disabled connectors to avoid unnecessary calls to the platform

andyh565:


packages/hull/src/middlewares/full-context-fetch.js
```diff
@@ -59,6 +60,26 @@ async function fetchSegments(ctx, entity = "user", cache) {
...
+    try {
+      debug(`Connector not found: ${error.message}`);
+    } catch (e2) {
+      debug("Error thrown in debug message");
+    }
+    return next(new ConnectorNotFoundError("Invalid id / secret"));
+  }
+  if (error.status === 402) {
+    try {
+      debug(`Organization is disabled: ${error.message}`);
```
sbellity, line 74: with the volumes we are getting on those endpoints I think it would be wiser to have some kind of metric instead of logging specifically here

PR #579, opened 2020-05-17
Feature/processor warnings

unity:
- Support new helper methods (`setIfNull(value)`, `increment(value)`, `decrement(value)`) 
- Adds a warning if using `anonymous_id` in attributes
- Adds support for `Optional Chaining`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining in backend code


packages/connectors/hull-processor/assets/readme.md
```diff
@@ -535,6 +535,7 @@ The processor provides the following methods to help you:
...
 | `leftSegment(<name>)`                              | Returns the segment object if the user just left the segment with the specified name; otherwise `null`. Please note that the name is case-sensitive.                                                                                                                                                  |
 | `leftAccountSegment(<name>)`                              | Returns the segment object if the user's Account just left the segment with the specified name; otherwise `null`. Please note that the name is case-sensitive.                                                                                                                                                  |
+| `setIfNull(value)`                              | A Helper to send a `setIfNull` command for the specified attribute. Use like this: hull.traits({ foo: setIfNull(1234) }).                                                                                                                                                  |
```
louisjahn, line 539: Could we have some description for the "increment" and "decrement" functions there? LGTM for the rest

michaloo:
I have created a story for it. Do we need to add `Optional Chaining` here?

packages/connectors/hull-processor/assets/readme.md
unity: I left them undocumented for now. They are here for completeness but I’m not sure they are useful to customers so I went for leaving them undocumented

PR #595, opened 2020-06-01
get status of member on incoming webhook; handle archived incoming webhook

andyh565:


andyh565:
@SMK1085 @michaloo the limitation here is that if a user is archived via the API, we will not receive a webhook for that action

packages/connectors/hull-mailchimp/server/actions/webhook.js
```diff
@@ -56,7 +54,8 @@ async function handleAction(
     case "unsubscribe":
       processedData = _.merge({}, data, {
         status: "unsubscribed",
-        subscribed: false
+        subscribed: false,
+        archived: data.action === "archive" || data.action === "delete"
```
andyh565, line 58: @SMK1085 what would you want the archived attribute to be? Is this fine?

SMK1085: @andyh565 the naming LGTM 👍

PR #596, opened 2020-06-01
Release 0.1.26

andyh565:


packages/connectors/hull-mailchimp/server/lib/sync-agent.js
```diff
@@ -317,6 +330,7 @@ class SyncAgent {
           if (_.get(envelope, "mailchimpCurrentMember.status")) {
             const asUser = this.client.asUser(envelope.message.user);
             return asUser.traits({
+              "mailchimp/archived": false,
```
michaloo, line 333: not sure if I understand that, why we are forcing `mailchimp/archived` to `false` here.
We are sending users with `status_if_new` which should not overwrite status so the user could possibly be archived and could stay archived at this stage.

PR #604, opened 2020-06-04
Paginate when paging through forms list

louisjahn:


packages/connectors/hull-typeform/manifest.json
```diff
@@ -66,7 +66,8 @@
         "showOverwriteToggle": false,
         "placeholder": "Typeform Field",
         "allowCreate": true,
-        "loadOptions": "/schema/fields"
+        "loadOptions": "/schema/fields",
+        "source": "typeform"
```
louisjahn, line 70: Added this on top of the pagination, otherwise dashboard is creating typeform attributes without the "typeform/" prefix

PR #606, opened 2020-06-04
release/typeform-form-paging

louisjahn:


packages/connectors/hull-typeform/server/lib/service-client.js
```diff
@@ -82,8 +82,21 @@ class ServiceClient {
       .set("Authorization", `Bearer ${accessToken}`);
   }
 
-  getForms(): Promise<TypeformGetFormsResponse> {
-    return this.agent.get("/forms");
+  async getForms(): Promise<TypeformGetFormsResponse> {
```
michaloo, line 85: @louisjahn do we want to change the flow type or we ignore it? ;)

louisjahn: In this page: https://eslint.org/docs/rules/no-await-in-loop
In the "When Not To Use It" section, it says "the output of one iteration might be used as the input to another". 

I'm using this because I can't start another request if the previous one returned me an empty array.

michaloo: I think that sounds good, I was more referring that the flow type `Promise<TypeformGetFormsResponse>` is not true anymore.

louisjahn: You are right, let me update that

PR #634, opened 2020-06-16
warn user if using alias incorrectly

andyh565:


unity:
@andy This one too should be merged no ? https://github.com/hull/hull-connectors/pull/579

PR #646, opened 2020-06-24
Log users when sending user updates to mailchimp, and during error

louisjahn:


packages/connectors/hull-mailchimp/server/lib/sync-agent.js
```diff
@@ -361,17 +361,20 @@ class SyncAgent {
...
             } else if (envelope.temporaryError) {
               asUser.logger.error("outgoing.user.error", {
-                error: envelope.temporaryError
+                error: envelope.temporaryError,
+                user: envelope.message.user
               });
               res.errors += 1;
             } else {
               asUser.logger.info("outgoing.user.success", {
-                member: envelope.mailchimpNewMember
```
michaloo, line 374: I think this one was good. We need to have MailChimp contact info there, so the JSON object we just sent or attempted to sent to Mailchimp API. We don't log the whole original Hull User profile on any other connector as far as I know.

michaloo: so I would expect to have `member: envelope.mailchimpNewMember` on every `outgoing.user.*` logline. Does it make sense?

PR #659, opened 2020-06-30
Hotfix/vm linting

unity:
Make sure the Linter relies on the actual libraries we expose in the processors (still incomplete but way better than before)


unity:
@louisjahn @michaloo No love for this one? It fixes a `Warning` that sven has for a connector that uses `uuid`

PR #676, opened 2020-07-07
Lint

unity:
Linting

### DIFF
--- a/packages/hull/src/utils/extract-stream.js
+++ b/packages/hull/src/utils/extract-stream.js
@@ -24,13 +24,20 @@ const promiseToWritableStream = require("./promise-to-writable-stream");
  * @return {Promise}
  */
 function extractStream({
-  body,
+  url,
+  format,
   batchSize,
-  callback,
+  onData,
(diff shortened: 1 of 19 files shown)


packages/hull/src/utils/extract-stream.js
```diff
@@ -24,13 +24,20 @@ const promiseToWritableStream = require("./promise-to-writable-stream");
  * @return {Promise}
  */
 function extractStream({
-  body,
```
michaloo, line 27: @unity is it a breaking change?
`body.url -> url`
`body.format -> format`
?

unity: Indeed, forgot to commit the related change on processing-middleware. just added it

michaloo: was that the only place `extractStream` was used?
Just trying to make sure we can proceed with the PR, since this is not pure linting change.

unity: yes it was the only place

PR #715, opened 2020-07-28
Try catch in account update

louisjahn:


packages/connectors/hull-hubspot/server/handlers/account-update.js
```diff
@@ -7,7 +7,13 @@ export default async (
...
   messages: Array<HullAccountUpdateMessage>
 ) => {
-  const syncAgent = new SyncAgent(ctx);
-  await syncAgent.sendAccountUpdateMessages(messages);
-  return {};
+  try {
+    const syncAgent = new SyncAgent(ctx);
+    await syncAgent.sendAccountUpdateMessages(messages);
+    return {};
+  } catch (err) {
```
unity, line 14: You shouldn't "swallow" the error entirely here. what happenes when sendAccountUpdateMessage throws an error? It is supposed to bubble up to the handler which should send "retry" without any additional code. Not what is happening @louisjahn ?

andyh565: @unity @louisjahn the errors on outgoing accounts are logged inside the sendAccountUpdateMessages function. If the entire job fails, then, in this function sendAccountUpdateMessages, the error is handled and logged here:

`
...
    } catch (error) {
      this.hullClient.logger.error("outgoing.job.error", {
        error: error.message
      });
      return Promise.reject(error);
    }
...
`

and then the account handler will send the retry response. 

It looks like it's handled the same way for user update here: https://github.com/hull/hull-connectors/blob/master/packages/connectors/hull-hubspot/server/handlers/user-update.js

...which may or may not be correct, in which case we can update both the user and account handlers

PR #724, opened 2020-08-10
feature/google-sheets-events

louisjahn:


packages/connectors/hull-google-sheets/src/sidebar/mapping.jsx
```diff
@@ -42,16 +42,19 @@ class Mapping extends Component<Props, State> {
...
       googleColumns = [],
       hullAttributes = []
     } = this.props;
+    const description = "Pick names of attributes or create new ones.";
+    const eventDescription = "Map columns to event properties";
     return (
       <Fragment>
         <h4>Columns mapping</h4>
-        <p>Pick names of attributes or create new ones.</p>
+        <p>{type === "user_event" ? eventDescription : description}</p>
```
unity, line 57: You should just do `type === "user_event" ? "String A" : "String B"`. No need for additional variable declarations

packages/connectors/hull-google-sheets/src/sidebar/source.jsx
```diff
@@ -30,9 +32,9 @@ class Source extends Component<Props, State> {
   }
 
   render() {
-    const { source, sources } = this.props;
+    const { source, sources, type } = this.props;
     const invalidChars = hasInvalidCharacters(source);
-    return (
+    return type === "user_event" ? null : (
```
unity, line 37: In the future we can allow to define source for events too - it's stored in Hull, along with event "type"

packages/connectors/hull-google-sheets/types.js
```diff
@@ -55,4 +55,10 @@ export type AccountClaims = {
...
 };
 
-export type ClaimsType = UserClaims | AccountClaims;
+export type EventType = {
+  event_id?: string,
+  event_name?: string,
+  created_at?: string
+};
+
+export type ClaimsType = UserClaims | AccountClaims | EventType;
```
unity, line 64: Doesn't seem to match what we are storing. Makes the data schema very messy. Creates tech debt. I recommend to avoid.

packages/connectors/hull-google-sheets/src/sidebar/index.jsx
```diff
@@ -424,14 +445,30 @@ export default class Sidebar extends Component<Props, State> {
...
         />
+        {type === "user_event" && (
+          <EventSetup
+            valid={true}
+            type={type}
+            onChangeRow={this.handleChangeClaim}
+            googleColumns={googleColumns}
+            claims={this.getClaims()}
+            errors={
+              !isValidEventSetup(this.getClaims()) && [
```
unity, line 456: Isn't created_at mandatory too?

PR #744, opened 2020-08-17
bigquery mvp

andyh565:


packages/hull-connector-framework/src/purplefusion/auth/auth-utils.js
```diff
@@ -19,6 +20,8 @@ function getServiceOAuthParams(manifest, serviceDefinitions) {
       _.merge(params, oauth2);
     } else if (authentication.strategy === "hubspotoauth") {
       _.merge(params, hubspotOAuth);
+    } else if (authentication.strategy === "googleoauth") {
```
unity, line 23: That's ugly. Can't we have something less custom ?

packages/hull-connector-framework/src/purplefusion/auth/google-oauth.js
```diff
@@ -0,0 +1,80 @@
...
+  HullOAuthStatusResponse,
+  HullOAuthHandlerParams,
+  HullOauthAuthorizeMessage,
+  HullOAuthAuthorizeResponse
+} from "hull";
+import { cacheClientCredentials } from "../webhooks/cache-client-credentials";
+
+const _ = require("lodash");
+
+const googleOAuth = {
```
unity, line 14: Having connector specific code inside of the framework is a big anti pattern. Please find a way to have connector code inside connectors and not have to do this inside of the framework.

packages/hull-connector-framework/src/purplefusion/auth/auth-utils.js
```diff
@@ -5,6 +5,7 @@ const _ = require("lodash");
 const { isUndefinedOrNull } = require("../utils");
 
 const { hubspotOAuth } = require("./hubspot-oauth");
+const { googleOAuth } = require("./google-oauth");
```
unity, line 8: If you can't find a way to extract this from the framework I will have a look next week. We can't leave this like this. It's an anti pattern.

louisjahn:
We have planned for a rework of this oauth implementation in purple-fusion for our next sprint. We will be temporarily using this solution for this mvp.

louisjahn (approved):

andyh565:
> We have planned for a rework of this oauth implementation in purple-fusion for our next sprint. We will be temporarily using this solution for this mvp.

@louisjahn the googleoauth strategy that is being imported isn't being used. Can you remove that from bigquery/package.json and revert yarn.lock

PR #797, opened 2020-10-30
Feature/cluster

unity:
This branch has
- Merged `release/aws` and `master` - Should deploy both on Heroku and AWS
- Node V12
- Cluster mode with PM2
- Adds support for Lightweight Connectors
- Adds Dropcontact connector
- Adds Phantombuster connector
- Webpack as a separate that proxies the backend. Accelerates dev. cycle since front-end doesn't rebuild at every change.
- Updated eslint options
- Declarative Queuing
- Environment-driven defaults, with stripped down `config.js` files. No more lengthy `process.env` in `config.js`
- No more `server/index.js` for connectors - Loads `config.js` automatically
- No more `manifest.json` in `config.js` - Fetched automatically


unity:
Woohooo!

Romain Dardour, CEO @ hull.io
Schedule some time with me

PR #832, opened 2020-12-09
Feature/sql bull imports

unity:


unity:
@michaloo @andyh565 A quick test shows imports going through for PostgreSQL - must run complete tests though.

andyh565:
> @michaloo @andyh565 A quick test shows imports going through for PostgreSQL - must run complete tests though.
> Not sure if there was a reason to use SQS instead of a Redis Queue here.

@unity overall, seems good to me. @michaloo can go into further detail regarding the move from kue to SQS. There were issues using with it in the past. What isn't clear to me is whether the issue was kue or a redis based queue.

unity:
We probably were using Kue at the time.
Also @andyh565 I only gave a quick test to PostgreSQL - consider this PR a very early draft that just lays the basics.

michaloo:
Yeah, initially we were using kue and the problem was that workers were constantly getting stuck. Then we moved to bull and it didn't help, that's why we moved to SQS as a final try to resolve the problem.

PR #843, opened 2021-01-07
feature/bigquery-adapter

louisjahn:


unity:
Whoa. Such speed!  🚀 amazing how fast you turned this around @louisjahn ! 🙌😻

packages/hull-sql/server/lib/sync-agent.js
```diff
@@ -544,7 +544,7 @@ export default class SyncAgent {
       }
 
       const record = this.adapter.in.transformRecord
-        ? this.adapter.in.transformRecord(rawRecord)
+        ? this.adapter.in.transformRecord(rawRecord, this.ship.private_settings)
```
louisjahn, line 547: @andyh565 Is that alright if I change the function signature for this bigquery adapter?
There's an option to add a prefix to each attribute (like `bigquery/`) and that's stored in the private_settings.

andyh565: yeah that's fine

PR #848, opened 2021-01-12
Initial version of hunter connector

michaloo:


packages/connectors/hull-hunter/server/index.js
```diff
@@ -0,0 +1,84 @@
...
+  if (!payload['domain'] && !payload['company']) {
+    throw new SkippableError("Domain or company name not present on this user.");
+  }
+
+  const response = await request
+    .get("https://api.hunter.io/v2/email-finder")
+    .query(payload);
+
+  if (response.body && response.body.data.email) {
+    if (response.body.data.score > 90) {
```
unity, line 67: @michaloo should we make this a setting ? Not a fan of magic values.

unity:
Nice work! Looks amazing

packages/connectors/hull-hunter/server/index.js
```diff
@@ -0,0 +1,84 @@
...
+        email: response.body.data.email
+      };
+      const traits = {
+        "hunter/enriched_at": moment().format(),
+        "hunter/email": response.body.data.email,
+        "hunter/score": response.body.data.score
+      };
+      return client.asUser(userClaims).traits(traits);
+    }
+    throw new SkippableError(`Hunter.io returned score below 90: ${response.body.data.score}, ignoring results.`);
```
unity, line 80: Should we return data / status instead of throwing ?

unity (commented):
Comments made

packages/connectors/hull-hunter/server/index.js
michaloo: Dev part of me agrees with you 100%, PO part of me said: "deliver asap to the customer to get feedback".
I will make it configurable in the next iteration.

packages/connectors/hull-hunter/server/index.js
michaloo: hm, okay, I can have another look at what would happen exactly, it wasn't straightforward what logline would be produced in the end. I think if I just return it will do outgoing.user.success.

PR #855, opened 2021-01-21
updated description to be more accurate

timliuhull:
Based on what I read here: 

https://github.com/hull/hull-connectors/blob/hotfix/update-clearbit-doc/packages/connectors/hull-clearbit/server/clearbit/reveal.js#L37

If the segment is empty we send no one. Not sure how we're "Not sending anyone with identifiable information" either... but must be in there somewhere... How should this actually be setup?

### DIFF
--- a/packages/connectors/hull-clearbit/manifest.json
+++ b/packages/connectors/hull-clearbit/manifest.json
@@ -1497,7 +1497,7 @@
     },
     {
       "title": "User Reveal whitelist",
-      "description": "Who will be sent to Reveal. *Every anonymous user is sent if this is empty*. We never send anyone who has identifiable information",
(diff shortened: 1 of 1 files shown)


unity (approved):
Correct 👍 .

`Not sending anyone with identifiable information` is a statement from the previous connector.
This was changed to allow anyone in the reveal segment to be sent, following a past customer request. Docs need to be updated

timliuhull:
@unity What was the purpose of "Not sending anyone with identifiable information"? That sounds like maybe I should setup some blacklist for people with email maybe? Or people who already have an account?

unity:
It was an early implementation detail that ensured we didn't reveal people after we had some information about them, Early decision. Include and exclude lists are better suited.

PR #868, opened 2021-02-05
[hull-warehouse] user/account outgoing success logging

andyh565:


andyh565:
TODO fix logging in case where transformation does not take place (send_all_attributes is turned on)

PR #874, opened 2021-02-12
[hull-salesforce] auth flow

andyh565:


JoshQuiambao (approved):

packages/hull/src/handlers/oauth-handler/factory.js
```diff
@@ -89,44 +91,52 @@ function OAuthHandlerFactory({
...
-    ) {
-      if (done === undefined) {
-        done = profile;
-        profile = verifyParams;
-        verifyParams = undefined;
+  const { authorizationURL, tokenURL, name: strategyName } = strategy;
+
+  function authorize(req: HullRequest, res: HullResponse, next: NextFunction) {
+    const oauth_url = req.hull.connector.private_settings[
+      `${_.toLower(name)}_oauth_url`
```
unity, line 98: Instead of re-building a setting name here, which relies on the developer being diligent in declaring the setting, you could directly pass a direct reference, which is what we have done for `settings_sections` for instance.

It's more explicit, as there is no string manipulation and no convention to know about, and the server can later analyze to check if reference is valid and data exists this without having to know anything about the code.

Also I wouldn't shadow the `tokenURL` and `authorizationURL` with a computed value. Splitting the parts and exposing them as different keys increases clarity I think.

Lastly, if as we're spreading an object (`strategy`), I wouldn't include properties that aren't meant to be there inside, just to avoid potential clashes and conflicts.

I'd move the un-included settings outside.

As a result:

manifest.json
```json
"options": {
  "hostname": "private_settings. oauth_hostname",
  "token_path":  "/services/oauth2/token",
  "authorization_path": "/services/oauth2/authorize",
  "strategy": {
    "scope": ["refresh_token", "api"]
  }
}
```

factory.js
```js
const  { hostname, authorization_path , token_path } = opts;
...

const OAuthStrategy = new Strategy({
  ...strategy,
  authorizationURL: `${hostname}/${authorization_path}`,
  tokenURL:  @`${hostname}/${token_path}`,
}

```

unity (changes requested):

packages/hull/src/handlers/oauth-handler/factory.js
andyh565: thanks @unity! I'll apply these suggestions for the story I created here: https://www.notion.so/hull/hull-connectors-implement-oauth-to-have-configurable-base-url-572b4898579a45edb07838a87ac22767

PR #886, opened 2021-02-17
make ipCheck a more generic feature

unity:


andyh565:
> I actually got `undefined` as a response here @andyh565 - I don't know if results differ based on environement

hmm that's interesting. I saw the circleci test fail this way and I also ran the test locally to confirm the failure

andyh565 (approved):

